Issue #3, opened 2022-11-29
C++ <-> Python conversion performance


benbovy:
I did a quick benchmark to evaluate the overhead of C++ -> Python conversion:

```cpp
py::array_t<PyObjectGeography> create(py::array_t<double> xs, py::array_t<double> ys) {
    py::buffer_info xbuf = xs.request(), ybuf = ys.request();

    auto result = py::array_t<PyObjectGeography>(xbuf.size);
    py::buffer_info rbuf = result.request();

    double *xptr = static_cast<double *>(xbuf.ptr);
    double *yptr = static_cast<double *>(ybuf.ptr);
    py::object *rptr = static_cast<py::object *>(rbuf.ptr);

    py::gil_scoped_release();
    for (size_t i = 0; i < xbuf.shape[0]; i++) {
        auto point_ptr = PointFactory::FromLatLonDegrees(xptr[i], yptr[i]);
        
        // "cast" benchmark (calls py::cast)
        rptr[i] = py::cast(std::move(point_ptr));
        // or
        // "no_cast" benchmark (just create an empty python object)
        rptr[i] = py::object()
    }

    return result;

...

m.def("create", &create);
```

Here are the results:

```python
x = np.random.rand(10_000)
y = np.random.rand(10_000)

# "cast" benchmark
%timeit s2shapely.create(x, y)
# 12.6 ms ± 241 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# "no cast" benchmark
%timeit s2shapely.create(x, y)   
# 4.46 ms ± 156 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The cast version is almost 3x slower than the no cast version. That's a lot for "just" wrapping the C++ geography in a Python object. Not sure what's happening (the Geography (sub)classes only implement move semantics, so there shouldn't be any data copy at this level), but there must be something wrong.

benbovy:
Tracked this down in pybind11's `type_caster_generic::cast`.

The bottleneck seems that pybind11 keeps a map of all registered instances and first tries to find and return the corresponding instance from the map, creating a new one if not found.

In the example above (more generally in geography creation functions), we know that no instance already exists, we could skip this step.

benbovy:
Now a quick benchmark to evaluate the overhead of Python -> C++ conversion:

```cpp
int dummy(PyObjectGeography obj) {
    // equivalent to `obj.cast<Geography*>()` where obj is a `py::object`
[...]

PR #5, opened 2022-12-01
Optimize Python -> C++ Geography conversion

benbovy:
Partially improves #3

It is not very nice to rely on pybind11 internals but I'm afraid that there's currently no way around that. It's probably worth it given the large gain in performance over directly using `pybind11::cast()` (almost 10x).

Not sure if it is really safe to use a static class member to store the Geography python type pointer as a cache, but it seems to work. I'll wait adding CI configuration (run tests) in another PR before merging this.

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -1,0 +1,112 @@
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+name: Run tests
+
+jobs:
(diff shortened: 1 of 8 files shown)


benbovy:
> Not sure if it is really safe to use a static class member to store the Geography python type pointer as a cache

Maybe a cleaner solution would be to store the type in a capsule, e.g., via some init function called just after importing the compiled Python extension (would require installing it as a subpackage).

PR #9, opened 2022-12-16
Add documentation

benbovy:


benbovy:
I'm going to merge this. I added an API reference section with docstrings for all API implemented so far. I plan to add some installation instructions + other useful details in a follow-up PR.

src/geography.cpp
```diff
@@ -133,26 +137,78 @@ PyObjectGeography destroy_prepared(PyObjectGeography obj) {
...
+                                      R"pbdoc(
+        Returns the inherent dimensionality of a geometry.
+
+        The inherent dimension is 0 for points, 1 for linestrings and 2 for
+        polygons. For geometrycollections it is the max of the containing elements.
+        Empty collections and None values return -1.
+
+    )pbdoc");
+
+    pygeography.def_property_readonly("nshape", &Geography::num_shapes, R"pbdoc(
```
jorisvandenbossche, line 168: (not directly related to this PR, just thinking about it while looking through this)

We could also keep the `num_shapes` naming on the python side? (to keep it consistent with c++, and I also find it a bit more readable, and both are different from shapely anyway)

src/geography.cpp
```diff
@@ -133,26 +137,78 @@ PyObjectGeography destroy_prepared(PyObjectGeography obj) {
...
+
+    pygeography.def("__repr__", [](const Geography &geog) {
+        s2geog::WKTWriter writer;
+        return writer.write_feature(geog.geog());
+    });
 
-    py::class_<Point, Geography>(m, "Point")
-        .def(py::init(&PointFactory::FromLatLonDegrees));
+    auto pypoint = py::class_<Point, Geography>(m, "Point", R"pbdoc(
+        A geography type that represents a single coordinate with lat,lon or x,y,z values.
```
jorisvandenbossche, line 180: Why the "lat, lon or x, y, z"? (z-coordinate doesn't seem to be supported by s2?)

src/geography.cpp
benbovy: Agreed.

src/geography.cpp
benbovy: I copied the docstrings from shapely but indeed it is not yet implemented here. It is possible to define points using x,y,z geocentric coordinates (on the unit sphere) in s2geometry (I think it is also the model used internally). We could support it here although it has not the same meaning than the z coordinate in GEOS?

PR #12, opened 2022-12-18
Add instructions for setting up local development environment

jorisvandenbossche:
This is of course quite specific and opinionated, but on the short term I think that's fine? This should at least make it a bit easier to quickly set up for people using conda.


benbovy:
Great, thanks @jorisvandenbossche. Yes this is certainly better than the previous installation instructions!

PR #13, opened 2022-12-19
Add logo, readme and doc tweaks

benbovy:
A brand new logo (based on S2 cells) + a few minor changes in Readme and docs (changed the theme to sphinx-book-theme).


jorisvandenbossche:
Nice ;) 
I do like the logo, two comments: 1) technical nit, the svg file seems truncated? (the last part of "spherely" is cut off, here on github but also in a local viewer), and 2) the typical feedback will be that it has to many small details when the logo is shown in a smaller format?

benbovy:
Thanks for the feedback @jorisvandenbossche :)

> 1) technical nit, the svg file seems truncated? (the last part of "spherely" is cut off, here on github but also in a local viewer)

Hmm that's weird? It looks good in my browser..

Screenshot 2022-12-20 at 10 23 14
Screenshot 2022-12-20 at 10 23 38

> the typical feedback will be that it has to many small details when the logo is shown in a smaller format?

Yes I agree there's a lot of details (which also makes the svg file pretty heavy for a logo). Maybe we could remove the S2 cells thin boundaries? I kind of like it, though.

benbovy:
The preview on MacOS is indeed truncated.

martinfleis:
We should also have a png versions. Those can also then be used in docs of svg proves tricky. 

I think that in this case the level of detail is okay as the logo itself seems to work when tiny as well. You just see the large shapes which is okay to me.

benbovy:
I've updated multiple versions of the logo (with / without cell border lines) that can be used depending on the size. I also added png exports + a favicon.

spherely_logo_large

spherely_logo_noline_medium

favicon copy

martinfleis (approved):
Looks good to me! 

Btw, I don't see it truncated on macOS in Safari.

benbovy:
Ok, thanks! Merging. (it is truncated when seeing the svg file with MacOS's Finder preview -- maybe an issue with svg thumbnails? -- but it looks good in web browser indeed).

jorisvandenbossche:
I have the same issue with the cut-off image in the README:

image


But so this might be something Firefox specific (or Firefox+linux), since in chrome it looks better. Still, it would be nice to fix this (there are still _some_ people that use Firefox ;))

martinfleis:
I guess because the text is still as a text and you don't have the font instead of the text being in curves. We should fix that.

benbovy:
It looks indeed like a font issue, hopefully fixed in [f86f6a0](https://github.com/benbovy/spherely/commit/f86f6a0ff602274c580b3059bb7b861f3a439f13). @jorisvandenbossche does it look good now?

Thanks for raising this.

> there are still some people that use Firefox ;)

Including me :-)

jorisvandenbossche:
Indeed, that works fine now! (and also much better looking with a nice font instead of that serif font ;))

PR #15, opened 2023-02-24
Add Polygon

benbovy:
Also add the possibility to create a `LineString` using `Point` objects.

`Polygon` has currently a few limitations that we can address later:

- Only shell vertices can be provided (no holes)
- The order of vertices is always normalized (in S2Geometry a `S2Loop` is oriented CCW in order to unambiguously define its interior but S2Geometry provides a method to automatically normalize the order by checking whether the area inside the loop is greater than 1/2 sphere). For performance reasons we might want to expose an option to skip this normalization.
- The validity of the polygon (loop) is always checked, but again sometimes we might want to skip this check for performance reasons.

S2Geography actually provides constructor classes with such `check` and `oriented` options. Maybe we can simply reuse it, although these classes return a unique pointer to the Geography base class which may make things a bit more complicated on the spherely side.


src/geography.hpp
```diff
@@ -75,16 +76,31 @@ class Point : public Geography {
...
+        return GeographyType::LineString;
+    }
+};
+
+class Polygon : public Geography {
+public:
+    Polygon(S2GeographyPtr&& geog_ptr) : Geography(std::move(geog_ptr)){};
+
+    inline GeographyType geog_type() const override {
         return GeographyType::LineString;
```
jorisvandenbossche, line 104:
```suggestion
        return GeographyType::Polygon;
```

tests/test_geography.py
```diff
@@ -11,13 +11,58 @@ def test_point() -> None:
...
+    poly = spherely.Polygon(coords)
+    assert poly.dimensions == 2
+    assert poly.nshape == 1
+    assert repr(poly).startswith("POLYGON ((0 0")
+
+
+def test_polygon_error() -> None:
+    with pytest.raises(ValueError, match="loop is not valid.*duplicate vertex.*"):
+        # polygon vertices should be open (duplicate vertex error)
+        spherely.Polygon([(0, 0), (0, 2), (2, 0), (0, 0)])
```
jorisvandenbossche, line 45: Is that something we can control? Because that's different as in shapely, and eg also the WKT shows it as a closed ring (from my other branch using convex_hull, the object is just still a base class Geography object, but the repr already works):
```
In [16]: line = spherely.LineString([(0, 0), (2, 0), (2, 2)])

In [17]: res = spherely.convex_hull(line)

In [18]: res
Out[18]: POLYGON ((0 0, 2 2, 0 2, 0 0))
```

So I would expect that I can recreate such a Polygon from the coordinates shown in the repr

jorisvandenbossche (commented):
Nice!

src/geography.hpp
benbovy: Good catch, thanks!

jorisvandenbossche: Maybe add a polygon to the `test_get_type_id` test to cover this

tests/test_geography.py
jorisvandenbossche: Sounds good (was also thinking about that manual check on our side if this is not something that can be controlled otherwise)

jorisvandenbossche: And this can certainly be left for a follow-up I think, it will be good to have this basic polygon class for other PRs and writing proper tests

PR #16, opened 2023-03-08
use s2geography from conda

martinfleis:
Switching from building s2geography from source to getting it from conda.


martinfleis:
@benbovy this can be merged and then used in the other two PRs to check them against both stable and latest.

benbovy:
Excellent, thanks @martinfleis!

PR #18, opened 2023-03-08
add: some predicates

ccalvocm:
Add some predicates to spherely like within and disjoint.


src/predicates.cpp
```diff
@@ -73,4 +89,28 @@ void init_predicates(py::module& m) {
             Geography object(s)
 
     )pbdoc");
-}
+    
+    m.def("within", py::vectorize(&within), py::arg("a"), py::arg("b"),
+          R"pbdoc(
+        Returns True if A does not intersect at all with B.
```
benbovy, line 95:
```suggestion
        Returns True if A is completely inside B.
```

src/predicates.cpp
```diff
@@ -62,7 +78,7 @@ void init_predicates(py::module& m) {
 
     m.def("contains", py::vectorize(&contains), py::arg("a"), py::arg("b"),
           R"pbdoc(
-        Returns True if B is completely inside A.
+        Returns True if A is completely inside B.
```
benbovy, line 81: I think the original description is the correct one?

benbovy (commented):
Thanks a lot @ccalvocm! I have just two minor comments for the docstrings and then I think it's good to go.

benbovy:
@ccalvocm could you please add the two functions added in this PR in the documentation (API reference)? It's here: https://github.com/benbovy/spherely/blob/main/docs/api.rst#predicates.

Issue #8, opened 2022-12-15
Precision of the WKT repr


jorisvandenbossche:
It might be nice to get a better float (using some rounding) representation:

```
In [8]: spherely.LineString([(-1, -1), (1, 1)])
Out[8]: LINESTRING (-0.9999999999999998 -1, 0.9999999999999998 1)
```

paleolimbot:
It looks like I put this in the `WKTWriter` but didn't expose it: https://github.com/paleolimbot/s2geography/blob/master/src/s2geography/wkt-writer.cc#L20-L21

(The WKT writer is a shim and will be replaced by a new one that actually has dedicated tests: https://github.com/geoarrow/geoarrow-cpp/blob/main/src/geoarrow/geoarrow.h#L169-L184 )

jorisvandenbossche:
GEOS started to use https://github.com/ulfjack/ryu to use a natural "shortest" printing by default (eg 0.3 instead of 0.300000011..). Might be overkill on the short term, but just noting it here.

(that's also more an issue for s2geography)

paleolimbot:
Yes...also ~5x faster than snprintf (from my limited set of benchmarks)! It looks like I left a hook for it as a compile-time option ( https://github.com/paleolimbot/s2geography/blob/master/src/s2geography/wkt-writer.cc#L6-L11 ), although I'd like to replace this particular WKT writer with the one in geoarrow-cpp (which has exhaustive tests at the C level).

benbovy:
@jorisvandenbossche I think we can close this now?

jorisvandenbossche:
Indeed, we have an argument exposed to control the precision now, and have set the default for the repr to a lower value.

PR #20, opened 2023-03-08
Add centroid, boundary and convex_hull function (returning new Geography)

jorisvandenbossche:


src/accessors-geog.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "pybind11.hpp"
+
+namespace py = pybind11;
+namespace s2geog = s2geography;
+using namespace spherely;
+
+PyObjectGeography convex_hull(PyObjectGeography a) {
+    const auto& a_ptr = a.as_geog_ptr()->geog();
+    auto s2_obj = s2geog::s2_convex_hull(a_ptr);
+    auto geog_ptr = std::make_unique<Geography>(std::move(s2_obj));
```
jorisvandenbossche, line 13: So currently, this function returns instances of the base class `Geography`. If I change this to `std::make_unique<Polygon>`, then it actually returns a `Polygon` python subclass. 

For the convex_hull case, I think I can be _sure_ that this will always return a Polygon. But in general, do we need a method on `s2geography::Geography` to know which subtype it actually is (a PolygonGeography) in this case), so we can assert / known to which subclass to downcast on our side?

jorisvandenbossche: I don't know if this is the same for s2, but with GEOS there are certainly geometry-returning functions that can return varying geometry types depending on the input (eg an intersection can basically return any geometry type)

src/pybind11.hpp
```diff
@@ -90,6 +90,10 @@ class PyObjectGeography : public py::object {
         return py::cast(std::move(geog_ptr));
     }
 
+    // static PyObjectGeography as_py_object(std::unique_ptr<T> geog_ptr) {
+    //     return static_cast<PyObjectGeography>(py::cast(std::move(geog_ptr)));
+    // }
```
jorisvandenbossche, line 95: I need to clean this up, but I thought to have `as_py_object` actually return PyObjectGeography instead of py::object, since it's currently not yet used, and in practice we need PyObjectGeography in the vectorized functions. But didn't get this to work ..

src/accessors-geog.cpp
benbovy: > I don't know if this is the same for s2, but with GEOS there are certainly geometry-returning functions that can return varying geometry types depending on the input (eg an intersection can basically return any geometry type)

If you know the type of the input(s) can you always determine the type of the output(s), or is there some geometry-returning functions where the output geometry type depends on some internal processing logic?

In the 1st case, we could use `spherely::Geography::geog_type()` returned from the inputs. Although I agree this might be worth of having such method available in s2geography, which would help getting rid of `spherely::Geography` wrapper classes that makes things quite confused.

src/pybind11.hpp
benbovy: Something like this should work:

```cpp
template <class T, std::enable_if<...>>
static PyObjectGeography as_py_object(std::unique_ptr<T> geog_ptr) {
    auto pyobj = py::cast(std::move(geog_ptr));
    auto pyobj_geog = static_cast<PyObjectGeography&>(pyobj);
    return std::move(pyobj_geog);
}
```

src/accessors-geog.cpp
jorisvandenbossche: > If you know the type of the input(s) can you always determine the type of the output(s), or is there some geometry-returning functions where the output geometry type depends on some internal processing logic?

Yes, AFAIK that is the case. For example consider intersection of two polygons: if they overlap, you can get a Polygon or MultiPolygon, if they touch along a segment you get a LineString (or MultiLineString), and if they touch you get a Point (or MultiPoint), and if they both overlap in some area and touch in another area, you get a GeometryCollection.

src/pybind11.hpp
jorisvandenbossche: Thanks, will give that a try. Are you OK with changing the existing as_py_object to do that? (I am not sure if we have a use case of the current version that returns py::object)

benbovy: Yes sure, it is also more consistent to have a static method returning an instance of the same type IMO. (`from_geog` could be a better name than `as_py_object`).

tests/test_accessors.py
```diff
@@ -0,0 +1,83 @@
...
+            spherely.Point(1, 1),
+        ),
+    ],
+)
+def test_centroid(geog, expected) -> None:
+    # scalar
+    actual = spherely.centroid(geog)
+    assert isinstance(actual, spherely.Point)
+    # TODO add some way of testing almost equality
+    # assert spherely.equals(actual, expected)
```
jorisvandenbossche, line 27: Testing is not super easy at the moment (we will have to develop some utilities to help with this). 

Currently this gives failures like

```
>       assert spherely.equals(actual, expected)
E       assert False
E        +  where False = <built-in method equals of PyCapsule object at 0x7fea1738e790>(POINT (4.999999999999999 0), POINT (5 0))
E        +    where <built-in method equals of PyCapsule object at 0x7fea1738e790> = spherely.equals
```

So either we need to have a version of `equals` that does some rounding/snapping of coordinates (I think S2 should provide some options for that, which we probably want to expose). 
Or more generally test based on the WKT repr, but then we also need to have a way to specify / reduce the precision of the WKT repr (https://github.com/benbovy/spherely/issues/8)

src/accessors-geog.cpp
```diff
@@ -0,0 +1,72 @@
...
+    std::unique_ptr<Point> point =
+        make_geography<s2geog::PointGeography, spherely::Point>(s2_point);
+    return PyObjectGeography::from_geog(std::move(point));
+}
+
+PyObjectGeography boundary(PyObjectGeography a) {
+    const auto& a_ptr = a.as_geog_ptr()->geog();
+    auto s2_obj = s2geog::s2_boundary(a_ptr);
+    // TODO return specific subclass
+    auto geog_ptr = std::make_unique<spherely::Geography>(std::move(s2_obj));
```
jorisvandenbossche, line 22: Boundary can return various geography types (empty collection for point, multipoint for linestring, linestring for polygon), so for now using the generic `Geography`, which will also give this parent class at the python level. We will have to add some utility to infer the type and cast to correct subclass.

jorisvandenbossche:
@benbovy this should be ready I think. There are certainly some things to improve (easier testing, correct return class), but those can be left for follow-ups I think.

tests/test_accessors.py
benbovy: I agree we need better helpers or features for testing (getting object coordinates would be useful too). This is short-term priority I think.

benbovy (commented):
Nice, LGTM thanks @jorisvandenbossche!

Could you just add the new functions in the API reference, please? And then I think we can merge this right away.

PR #21, opened 2023-03-09
Add more instructions in README (dev builds)

benbovy:
@jorisvandenbossche you might be interested in this (scikit-build-core now supports cached builds).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,17 @@ This library is at an early stage of development.
 
 - Python
 - Numpy
-- s2geography
-- s2geometry
+- [s2geography](https://github.com/paleolimbot/s2geography)
+- [s2geometry](https://github.com/google/s2geometry)
+
+Additional requirements when building spherely from source:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +53,14 @@ $ cd spherely
 $ python -m pip install . -v --no-build-isolation
 ```
 
+Note that you can specify a build directory in order to avoid rebuilding the
+whole library from scratch each time after editing the code (requires
+scikit-build-core v0.2.0+):
+
+```
+$ python -m pip install . -v --no-build-isolation --config-settings build-dir=build/skbuild
```
jorisvandenbossche, line 61: Yes, I started to use this yesterday and was planning to open a PR to add it to the readme ;)

You can use `skbuild/{cache_tag}` to have builds for multiple python versions next to each other, but maybe most contributors won't need that.

jorisvandenbossche (approved):

PR #22, opened 2023-03-13
Predicates functor

benbovy:
This save some code duplication and will be slightly easier to refactor.

It also creates the `S2BooleanOperation::Options` object only once, although I don't expect any significant speedup and we might want to eventually expose options to Python (thus creating a new object for each call).


jorisvandenbossche (approved):
Nice!

PR #23, opened 2023-03-13
Add support for getting projected coordinates

jorisvandenbossche:
Just putting it here as draft PR to show what I was experimenting with (this just projects a single point at the moment). 

The reason I started looking at this is because it would be nice to get the coordinates in a certain projection for quick plotting of spherely geography objects (https://dewey.dunnington.ca/post/2022/s2-version-1.1.0/#plotting, https://github.com/r-spatial/s2/blob/76e36ae9ef1348412726e780f6a5a9e4ba0805d8/R/plot.R). 
But also in general we will want to expose methods to get all the coordinates, and then providing an option to get projected coordinates (in some basic projections that S2 supports) might be nice. 

For the plotting, we will also need to add tesselation in the process of getting the projected coordinates (to correctly represent the shortest-path-on-the-sphere in a projection that uses "straight" lines)

### DIFF
--- a/src/projections.cpp
+++ b/src/projections.cpp
@@ -1,0 +1,49 @@
+#include <stdexcept>
+#include <iostream>
+
+#include <s2/s2projections.h>
+#include <s2/s2latlng.h>
+#include <s2geography.h>
+
+#include "geography.hpp"
+#include "pybind11.hpp"
+
+namespace py = pybind11;
+namespace s2geog = s2geography;
+using namespace spherely;
+
+
+std::tuple<double, double> project_mercator(py::object obj) {
+    auto geog = (static_cast<PyObjectGeography&>(obj)).as_geog_ptr();
+    if (geog->geog_type() != GeographyType::Point) {
+        throw py::value_error("test");
+    }
+    auto point = static_cast<Point*>(geog);
+    auto s2point = point->s2point();
+
+    auto projection = S2::MercatorProjection(20037508.3427892);
+    auto r2point = projection.Project(s2point);
+    double x = r2point.x();
+    double y = r2point.y();
+
+    return std::make_tuple(x, y);
+}
+
+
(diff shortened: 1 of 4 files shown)


src/projections.cpp
```diff
@@ -0,0 +1,49 @@
...
+    //     Intersects implies that overlaps, touches and within are True.
+
+    //     Parameters
+    //     ----------
+    //     a, b : :py:class:`Geography` or array_like
+    //         Geography object(s)
+
+    // )pbdoc");
+
+    m.def("project_mercator", &project_mercator);
```
jorisvandenbossche, line 47: A problem with vectorizing a function like this is that it returns multiple values for each input value (so in numpy ufunc terms, it would be a generalize ufunc with a signature of "()->(n)" (adding one dimension)). But this is not something that `py::vectorize` supports AFAIK. So we might need to write this more manually (eg writing the loop ourselves, only supporting 1D input, which is what we do in shapely as well). 

And for non-point geographies, it doesn't fit in the ufunc paradigm anyway, since you have a variable number of coordinate pairs that gets returned (like `shapely.get_coordinates`)

benbovy:
Thanks for looking into this @jorisvandenbossche. I agree it would be nice to have an option to get coordinates in a given projection with support for tessellation.

src/projections.cpp
benbovy: Should we follow `shapely.get_coordinates` API then? I.e., return a (N, 2) array with a flat index. I guess with tessellation enabled we'll almost always want to get the index.

jorisvandenbossche:
Going to close this, as we now have projection support through either WKB/WKT or for coordinate values through geoarrow. 

We can still discuss whether we want something more similar to shapely`s `get_coordinates` for geographies in general (not just points), and at that point such function could also support projection/tessselation. But let's first discuss that in itself.

PR #24, opened 2023-03-14
More CI (mypy), setup pre-commit, formatting

benbovy:


docs/api.rst
```diff
@@ -36,7 +36,7 @@ Geography creation
 ------------------
 
 .. autosummary::
-   :toctree: _api_generated/
+   :toctree: _api_generated/e
```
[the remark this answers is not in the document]
benbovy: Yes indeed :)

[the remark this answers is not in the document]
benbovy: That was on purpose so I can test the cache featured in micromamba's GHA haha.

.clang-format
```diff
@@ -1,10 +1,17 @@
...
 PointerAlignment: Left
 ReferenceAlignment: Pointer
 IndentAccessModifiers: false
 AccessModifierOffset: -4
+BinPackArguments: false
+BinPackParameters: false
+ExperimentalAutoDetectBinPacking: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortLambdasOnASingleLine: Inline
```
jorisvandenbossche, line 16: I don't have _any_ opinion about how C++ code should look like, and how much we want to customize compared to the base Google style (so happy to follow your preferences). But just to consider, s2geography just follows google (https://github.com/paleolimbot/s2geography/blob/master/.clang-format), and arrow uses only minimal overrides (https://github.com/apache/arrow/blob/main/.clang-format)

 (but so +1 on adding a linter for this to enforce consistent style!)

benbovy: I don't have strong opinions either and I think it is a good idea to stay close to the Google style used and not add too much customization. This is just to fix some minor things I find annoying, like packing some of the parameters on the same line when all the parameters don't fit on one line, e.g., I prefer this:

```cpp
m.def("equals",
      py::vectorize(Predicate(s2geog::s2_equals)),
      py::arg("a"),
      py::arg("b"));  
```

over this:

```cpp
m.def("equals", py::vectorize(Predicate(s2geog::s2_equals)), py::arg("a"),
      py::arg("b"));  
```

benbovy:
Switching to micromamba make things even faster! Also the cache is a bit more reliable now (it seems to fix issues on Windows).

@jorisvandenbossche any other comment or suggestion before I merge this?

.clang-format
jorisvandenbossche: As mentioned, I don't mind, so just follow your preferences ;)

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          python -m pip install . -v --no-build-isolation
+
+      - name: Install mypy
+        run: |
+          python -m pip install 'mypy<0.990'
+
+      - name: Run mypy
+        run: |
+          python -m mypy --install-types --non-interactive
```
jorisvandenbossche, line 39: Shall we also run pre-commit in this lint build?

jorisvandenbossche: Like https://github.com/geopandas/geopandas/blob/571e0b15f61dfbf957e347d166118444da712afd/.github/workflows/tests.yaml#L22 or https://github.com/shapely/shapely/blob/bc62f989c8bb2de38f9762a301b7b02c416d38b9/.github/workflows/lint.yml#L16-L17

benbovy: I installed pre-commit as a github app for this repository.

jorisvandenbossche: Ah, yes, I see it now in the "checks" listing in the UI

PR #26, opened 2023-03-15
Add more Geography subclasses

benbovy:
- [x] `LinearRing`
  - There is no corresponding class in s2geography (yet) and no class in s2geometry that we can use directly, but it was relatively (hopefully) easy to expose such type based on `S2LaxClosedPolylineShape` (+ `S2Region` implementation copied from `S2Polyline`). I added here some code that we might consider moving to s2geography later if it's worth it.
  - The name "LinearRing" might be a bit misleading considering that the segments are geodesics... Should we keep this name for consistency with Shapely or choose another name?
- [x] `Polygon`
  - add support for holes
  - allow duplicate start/end vertices
- [x] `MultiPoint`
- [x] `MultiLineString`
- [ ] `MultiPolygon`
- [x] `GeometryCollection`

Also:

- [ ] Use various kinds of geometries elsewhere in the tests (predicates, etc.)

Let's save the following for later:

- support of empty - or full - geographies
- vectorized creation functions


jorisvandenbossche:
Maybe something to also discuss more generally: do we want / need all those subclasses?

benbovy:
> Maybe something to also discuss more generally: do we want / need all those subclasses?

Do you mean that alternatively we could expose only one `Geography` class (with a geometry type property) along with top-level creation functions for each kind of geometry?

I see two advantages of having subclasses:

- it is convenient for quick creation of a few simple features (perhaps more readable too)
- it is consistent with Shapely (is there any plan to eventually drop those subclasses in Shapely?)

One argument against subclasses is that it is a bit more code to maintain (maybe some duplicate logic here and there), but that's not much in my opinion.

benbovy:
I haven't checked if shapely copies the input geometries when creating a collection... Here that's a requirement here if we want to reuse `s2geography::GeographyCollection` (which owns its features).

I added a ``clone`` method to the spherely wrapper classes which makes it easier.  I think it would make things even easier if this method was available in the s2geography classes as well.

There are a few other things that we would need to port into the s2geography classes (also address #27) before we can get rid of spherely's wrapper classes (which I'd be very happy to see!). I'll wait a bit more before opening an issue in the s2geography repository with a clear and detailed proposal.

benbovy:
Closing. Superseded by #51.

PR #33, opened 2024-08-19
CI: temporarily pin conda s2geometry to 0.10

jorisvandenbossche:
Because latest s2geography on condda-forge is not yet compatible with the latest s2geometry 0.11 (https://github.com/paleolimbot/s2geography/pull/19 already fixed, but will take a bit until that is released and propagated to conda packages)


jorisvandenbossche:
The dev builds are still failing, but going to merge this to have the others passing

PR #34, opened 2024-08-19
BLD: require latest s2geometry 0.11 / s2geography 0.1.1 + use cmake config to find libraries

jorisvandenbossche:
If we require the latest release of both s2geometry and s2geography, that simplifies our CMake set-up a bit.

(this relies on having a conda-forge release)


benbovy:
s2geography last release is available on conda-forge (or should be in a few hours).

PR #35, opened 2024-08-19
CI: build sdist and wheels (Linux & MacOS)

benbovy:


.github/workflows/release.yml
```diff
@@ -0,0 +1,133 @@
...
+          path: ${{ runner.temp }}/3rd-party
+          key: ${{ matrix.os }}-${{ matrix.arch }}-${{ env.ABSL_VERSION }}-${{ env.S2GEOMETRY_VERSION }}-${{ env.S2GEOGRAPHY_VERSION }}-${{ hashFiles('ci/*') }}
+
+      - name: Fetch absl
+        if: steps.cache-build.outputs.cache-hit != 'true'
+        uses: actions/checkout@v4
+        with:
+          repository: abseil/abseil-cpp
+          ref: ${{ env.ABSL_VERSION }}
+          path: ${{ runner.temp }}/3rd-party/absl-src-${{ env.ABSL_VERSION }}
```
jorisvandenbossche, line 81:
```suggestion
          path: ${{ github.workspace }}/3rd-party/absl-src-${{ env.ABSL_VERSION }}
```

ci/install_3rdparty.sh
```diff
@@ -0,0 +1,115 @@
...
+    pip install cmake
+
+    rm -rf $BUILD_DIR/*
+    rm -rf $INSTALL_DIR/*
+
+    echo "Downloading, building and installing absl-$ABSL_VERSION"
+
+    cd $DEPENDENCIES_DIR
+    curl -OL https://github.com/abseil/abseil-cpp/archive/refs/tags/$ABSL_VERSION.tar.gz
+    tar xf abseil-cpp-$ABSL_VERSION.tar.gz -C $SRC_DIR
```
jorisvandenbossche, line 58:
```suggestion
    tar xf $ABSL_VERSION.tar.gz -C $SRC_DIR
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+            S2GEOGRAPHY_VERSION=${{ env.S2GEOGRAPHY_VERSION }}
+            CXX_STANDARD=${{ env.CXX_STANDARD }}
+          CIBW_BEFORE_ALL: ./ci/install_3rdparty.sh
+          CIBW_BEFORE_ALL_WINDOWS: ci\install_3rdparty.cmd
+          # CIBW_BEFORE_BUILD_WINDOWS: pip install delvewheel
+          CIBW_REPAIR_WHEEL_COMMAND_LINUX: 'LD_LIBRARY_PATH=/host${{ runner.temp }}/3rd-party/dist/lib64 auditwheel repair -w {dest_dir} {wheel}'
+          CIBW_REPAIR_WHEEL_COMMAND_MACOS: 'LD_LIBRARY_PATH=${{ runner.temp }}/3rd-party/dist/lib delocate-wheel --require-archs=${{ matrix.arch }} -w {dest_dir} -v {wheel}'
+          # CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: delvewheel repair --add-path ${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\bin -w {dest_dir} {wheel}
+          CIBW_TEST_REQUIRES: pytest
+          CIBW_TEST_COMMAND: pytest spherely
```
jorisvandenbossche, line 111:
```suggestion
          CIBW_TEST_COMMAND: pytest --pyargs spherely
```

benbovy: Hmm are the tests files installed with the spherely package?

jorisvandenbossche:
BTW, if you update 

https://github.com/benbovy/spherely/blob/38d3bf9df81e86140dacb962b573fc95e44f9316/pyproject.toml#L21-L21

to `>=3.10` then it should autmatically limit the wheels to those python versions (not that it will the error, but I just noticed it was building and testing a python 3.7 wheel, and we don't really need to provide wheels for that many python versions)

I bumped testing to use Python 3.10+ (https://github.com/benbovy/spherely/pull/36), so matching our requirements to that might be best.

.github/workflows/release.yml
jorisvandenbossche: Ah, no, not necessarily (probably not, actually, since they don't live inside the package). Didn't think about that, for shapely they are included and I know we struggled getting the correct `pytest` invocation in the cibuildwheel setup

jorisvandenbossche:
```suggestion
          CIBW_TEST_COMMAND: pytest tests
```

then ? (assuming we are in the cloned repo)

benbovy: Yes I guess we would either need to include the tests in the installed package or split into two different commands for linux (container) vs. macos so we can set the right path to the tests directory. Including tests may be better I guess.

jorisvandenbossche: Why would the path be different? When we are in the docker container, the cwd is not the repo source?

jorisvandenbossche: (given `./ci/install_3rdparty.sh` works to install the dependencies both on linux and mac, that makes me assume that `pytest tests/` should also work on both?)

benbovy:
```
libssl.3.dylib has a minimum target of 13.0
```

Looks like we'll need to build and install openssl manually on macos :-/

.github/workflows/release.yml
benbovy: Yes that's what I set in the last commit, let's see

jorisvandenbossche:
> ```
> libssl.3.dylib has a minimum target of 13.0
> ```
> 
> Looks like we'll need to build and install openssl manually on macos :-/

Could also (or at least for now) bump the deployment target (it just means the wheels are only installable on recent macos)

.github/workflows/release.yml
jorisvandenbossche: OK, so I was wrong :(

```
  ERROR: file or directory not found: tests
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+            S2GEOGRAPHY_VERSION=${{ env.S2GEOGRAPHY_VERSION }}
+            CXX_STANDARD=${{ env.CXX_STANDARD }}
+          CIBW_BEFORE_ALL: ./ci/install_3rdparty.sh
+          CIBW_BEFORE_ALL_WINDOWS: ci\install_3rdparty.cmd
+          # CIBW_BEFORE_BUILD_WINDOWS: pip install delvewheel
+          CIBW_REPAIR_WHEEL_COMMAND_LINUX: 'LD_LIBRARY_PATH=/host${{ runner.temp }}/3rd-party/dist/lib64 auditwheel repair -w {dest_dir} {wheel}'
+          CIBW_REPAIR_WHEEL_COMMAND_MACOS: 'DYLD_LIBRARY_PATH=${{ runner.temp }}/3rd-party/dist/lib delocate-wheel --require-archs=${{ matrix.arch }} -w {dest_dir} -v {wheel}'
+          # CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: delvewheel repair --add-path ${{ runner.temp }}\geos-${{ env.GEOS_VERSION }}\bin -w {dest_dir} {wheel}
+          CIBW_TEST_REQUIRES: pytest
+          CIBW_TEST_COMMAND: pytest tests
```
jorisvandenbossche, line 111:
```suggestion
          CIBW_TEST_COMMAND: pytest {project}/tests
```

jorisvandenbossche:
Success! Will have to add an upload step now to further verify that it actually works

benbovy:
> Success!

Yes finally!

We'll need to add windows too (maybe in a follow-up PR).

jorisvandenbossche:
Great!

PR #37, opened 2024-08-19
Add get_lat and get_lng vectorized functions (Point-only)

jorisvandenbossche:
Awaiting a more complex function to get coordinates in general, adding a simple get_lng/get_lat functions just for points (so those can easily be vectorized)

### DIFF
--- a/src/coords.cpp
+++ b/src/coords.cpp
@@ -1,0 +1,64 @@
+#include <s2/s2latlng.h>
+#include <s2geography.h>
+
+#include <iostream>
+#include <stdexcept>
+
+#include "geography.hpp"
+#include "pybind11.hpp"
+
+namespace py = pybind11;
+namespace s2geog = s2geography;
+using namespace spherely;
+
+double get_lat(PyObjectGeography a) {
+    auto geog = a.as_geog_ptr();
+    if (geog->geog_type() != GeographyType::Point) {
+        throw py::value_error("Only Point geometries supported");
+    }
+    auto point = static_cast<Point*>(geog);
+    auto s2point = point->s2point();
+    auto latlng = S2LatLng(std::move(s2point));
+    double lat = latlng.lat().degrees();
+    return lat;
+}
+
+double get_lng(PyObjectGeography a) {
+    auto geog = a.as_geog_ptr();
+    if (geog->geog_type() != GeographyType::Point) {
+        throw py::value_error("Only Point geometries supported");
+    }
+    auto point = static_cast<Point*>(geog);
+    auto s2point = point->s2point();
+    auto latlng = S2LatLng(std::move(s2point));
+    double lng = latlng.lng().degrees();
+    return lng;
+}
+
+void init_coords(py::module& m) {
+    m.def("get_lat",
+          py::vectorize(&get_lat),
+          py::arg("a"),
+          R"pbdoc(
+        Returns the latitude value of the Point (in degrees).
+
+        Parameters
+        ----------
+        a: :py:class:`Geography` or array_like
+            Geography object(s).
+
+    )pbdoc");
+
+    m.def("get_lng",
+          py::vectorize(&get_lng),
+          py::arg("a"),
+          R"pbdoc(
+        Returns the longitude value of the Point (in degrees).
+
+        Parameters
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
I see that R's `s2` (based on bigquery?) still calls this `st_x` and `st_y`, so should we also use `get_x` and `get_y` here?

martinfleis:
> should we also use get_x and get_y here?

+1 on that as that is also consistent with shapely.

src/coords.cpp
```diff
@@ -0,0 +1,64 @@
...
+double get_y(PyObjectGeography a) {
+    auto geog = a.as_geog_ptr();
+    if (geog->geog_type() != GeographyType::Point) {
+        throw py::value_error("Only Point geometries supported");
+    }
+    auto point = static_cast<Point*>(geog);
+    auto s2point = point->s2point();
+    auto latlng = S2LatLng(std::move(s2point));
+    double lat = latlng.lat().degrees();
+    return lat;
```
benbovy, line 23: We could also just call `s2geography::s2_y`.

src/spherely.pyi
```diff
@@ -107,6 +107,11 @@ contains: _VFunc_Nin2_Nout1[Literal["contains"], bool, bool]
 within: _VFunc_Nin2_Nout1[Literal["within"], bool, bool]
 disjoint: _VFunc_Nin2_Nout1[Literal["disjoint"], bool, bool]
 
+# coords
+
+get_y: _VFunc_Nin1_Nout1[Literal["get_lan"], float, np.float64]
```
benbovy, line 112:
```suggestion
get_y: _VFunc_Nin1_Nout1[Literal["get_y"], float, np.float64]
```

tests/test_coords.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import pytest
+
+import spherely
+
+
+def test_get_y_lng() -> None:
+    # scalar
+    a = spherely.Point(1.5, 2.6)
```
benbovy, line 9:
```suggestion
    a = spherely.point(1.5, 2.6)
```

(also for the other points created below)

tests/test_coords.py
```diff
@@ -0,0 +1,29 @@
...
+    actual = spherely.get_x(arr)
+    expected = np.array([1, 2, 3], dtype="float64")
+    np.testing.assert_allclose(actual, expected)
+
+    # only points are supported
+    with pytest.raises(ValueError):
+        spherely.get_y(spherely.LineString([(0, 1), (1, 2)]))
+
+    with pytest.raises(ValueError):
+        spherely.get_x(spherely.LineString([(0, 1), (1, 2)]))
```
benbovy, line 29:
```suggestion
    with pytest.raises(ValueError):
        spherely.get_y(spherely.linestring([(0, 1), (1, 2)]))

    with pytest.raises(ValueError):
        spherely.get_x(spherely.linestring([(0, 1), (1, 2)]))
```

benbovy (commented):
This is (almost) ready I guess?

Alternatively we could return NaN instead of returning an error for non-point geographies? Or expose a parameter that allows user choosing between the two behaviors (similarly to the `SAFE` prefix in bigquery)?

(also both functions can be added to the doc API reference).

src/coords.cpp
jorisvandenbossche: Yes, at the time of writing this PR I didn't realize that existed, in the meantime I had noticed that as well. 
I think that also has the behaviour of returning NaN for non-point geometries, as you suggested, if we would use it as is, without any validation before calling it.

I am not entirely following what the implementation is doing though: https://github.com/paleolimbot/s2geography/blob/5b06341403ec2e030b2a254bfa96d8c3386296f3/src/s2geography/accessors.cc#L165-L178 
When is it possible that you need that for loop in there? Or does that mean that if you have a collection of two points, it will just give you the x/y of the first point? (I would find that a bit unexpected behaviour)

benbovy: Yes I was a bit confused by the implementation too. Probably safer to keep raising an error in case of non-point geographies for now.

benbovy (approved):

Issue #17, opened 2023-03-08
Add bindings (top-level "ufuncs") for predicates, measurement, etc. 


benbovy:
Some features available in s2geography that could be implemented in spherely with relatively little effort (i.e., straightforward functions returning simple types like bool, float, etc.):

Predicates:

- [x] intersects
- [x] equals
- [x] contains
- [x] touches (#56)
- [x] within (#18)
- [x] disjoint (#18)
- [x] covers (#56)
- [x] covered_by (#56)

Measurement:

- [x] distance ([s2geography link](https://github.com/paleolimbot/s2geography/blob/495bffd5ef44b230cc120ab2634c98296e3b9dd8/src/s2geography/distance.h#L8)) (#43)
- [x] area ([s2geography link](https://github.com/paleolimbot/s2geography/blob/2b4a63e768047b1b7c59a5894beff082c02e1294/src/s2geography/accessors.cc#L112)) (#73)
- [x] length ([s2geography link](https://github.com/paleolimbot/s2geography/blob/2b4a63e768047b1b7c59a5894beff082c02e1294/src/s2geography/accessors.cc#L131)) (#77)

### One example with `intersects`

s2geography API:

https://github.com/paleolimbot/s2geography/blob/764682d60f757383e5853a45a12404c84f3e5961/src/s2geography/predicates.h#L10-L12

spherely bindings:

https://github.com/benbovy/spherely/blob/e874e6091888db6c09d9aee79c3ff260224f635b/src/predicates.cpp#L37-L51

For predicates that are not directly implemented in s2geography but that can be easily derived from another predicate, it is possible to use a lambda, like for `within`:

https://github.com/benbovy/spherely/blob/e874e6091888db6c09d9aee79c3ff260224f635b/src/predicates.cpp#L84-L100

Documentation:

https://github.com/benbovy/spherely/blob/main/docs/api.rst#predicates

Typing annotations:

https://github.com/benbovy/spherely/blob/0b7d3bbcce1dffbff9e283f6e15ff5c3497e26de/src/spherely.pyi#L104

PR #43, opened 2024-08-30
Add binding to calculate distance between two geometries

gahjelle:
See #17 

Adds `distance()` measurement. The distance is calculated in meters per default, based on a new `EARTH_RADIUS_METERS` constant. The radius can also be customized in the call to `distance()`.

### DIFF
--- a/src/accessors-geog.cpp
+++ b/src/accessors-geog.cpp
@@ -7,6 +7,8 @@ namespace py = pybind11;
 namespace s2geog = s2geography;
 using namespace spherely;
 
+const double EARTH_RADIUS_METERS = 6371.01 * 1000;
+
 PyObjectGeography centroid(PyObjectGeography a) {
     const auto& a_ptr = a.as_geog_ptr()->geog();
     auto s2_point = s2geog::s2_centroid(a_ptr);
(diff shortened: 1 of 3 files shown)


src/accessors-geog.cpp
```diff
@@ -69,4 +79,23 @@ void init_accessors(py::module& m) {
...
+          R"pbdoc(
+        Calculate the distance between two geographies.
+
+        Parameters
+        ----------
+        a : :py:class:`Geography` or array_like
+            Geography object
+        b : :py:class:`Geography` or array_like
+            Geography object
+        radius : float
```
benbovy, line 97:
```suggestion
        radius : float, optional
```

benbovy (approved):
Thanks @gahjelle ! Just one nit-picking suggestion for the docstrings otherwise it all looks good!

gahjelle:
Thanks @benbovy I've updated the docstring.

tests/test_accessors.py
```diff
@@ -1,8 +1,9 @@
 import numpy as np
+import pytest
 
 import spherely
 
-import pytest
+earth_radius_meters = 6_371_010
```
jorisvandenbossche, line 6: This is not actually used?

gahjelle: Ah, good catch @jorisvandenbossche 

Thanks, I've removed it.

benbovy:
Many thanks @gahjelle !

PR #45, opened 2024-09-03
CI release: activate windows builds

benbovy:


benbovy:
Yes finally wheels on Windows!

@jorisvandenbossche perhaps you may want to have a look here? There is still room for improving the way we build python wheels here, e.g.,

- on windows choose the s2geometry version to install (require switching to vcpkg manifest mode if I understand well IIUC - a bit silly this cannot be done using the command line)
- on macos / linux use vcpkg as well to install openssl, abseil and s2geometry (might allow using older macos deployment target? I haven't checked)

This can be done later, though (I already spent a lot of time on packaging).

ci/install_3rdparty.cmd
```diff
@@ -0,0 +1,112 @@
+:: Build and install absl, s2 and s2geography on posix systems.
```
jorisvandenbossche, line 1: No too important, but this header is not up to date

jorisvandenbossche:
Great!

Agreed that if it is working, this is more than good enough for now, and we can see to streamline the setup more in the future. 

> on macos / linux use vcpkg as well to install openssl, abseil and s2geometry (might allow using older macos deployment target? I haven't checked)

With pyogrio I have a good experience with using vcpkg for all platforms, so I would prefer that as well. Especially seeing the hoops you had to go through to get all dependencies properly installed in the end in https://github.com/benbovy/spherely/pull/35/, I was already wondering if vcpkg wouldn't make that easier (but that is not a given, of course .. ;))

One annoyance with the vcpkg setup is of course that s2geography is not included and so requires to still have some custom install script anyway. I don't know how easy it is to add a package to vcpkg upstream, but we could also have a local "overlay port" (https://learn.microsoft.com/en-us/vcpkg/concepts/overlay-ports), which might simplify to integrate s2geography in a full vcpkg workflow. 
Essentially that requires writing just a portfile.cmake for s2geography, and the one for s2geometry looks like this: https://github.com/microsoft/vcpkg/blob/master/ports/s2geometry/portfile.cmake (and I think for s2geography it would be very similar)

But all ideas for later ;)

benbovy:
Caching seems to work well.

jorisvandenbossche (approved):

benbovy:
Thanks Joris!

PR #50, opened 2024-10-04
ENH: add from_wkt/to_wkt functions

jorisvandenbossche:
This adds vectorized `from_wkt`/`to_wkt` functions. 

In addition to the geoarrow IO in https://github.com/benbovy/spherely/pull/49, this is still useful to have so users can convert to/from WKT with plain numpy arrays as well instead of having to go through Arrow (and use some arrow library).


src/io.cpp
```diff
@@ -0,0 +1,81 @@
...
+    options.set_oriented(oriented);
+    if (planar) {
+        auto tol = S1Angle::Radians(100.0 / EARTH_RADIUS_METERS);
+        options.set_tessellate_tolerance(tol);
+    }
+    s2geog::WKTReader reader(options);
+#else
+    if (planar || oriented) {
+        throw std::invalid_argument(
+            "planar and oriented options are only available with s2geography >= 0.2");
```
brendan-ward, line 25:
```suggestion
            "planar and oriented options are only available with s2geography >= 1.2");
```

(per above `if`?)

jorisvandenbossche: I _think_ I have it correct now.. In any case, the s2geography version is certainly 0.2 (https://github.com/paleolimbot/s2geography/blob/9ebd7b106ba2ae30046cb4f3e9561a8bbdb18b9b/CMakeLists.txt#L6).

I initially did the `#if` version check wrong, see https://github.com/benbovy/spherely/pull/50/commits/de19f0eb819b302f3a88d56d79e87f3b33b0c788 for correcting it. 

We want that both version of 0.2 or 1.1 result in True, with the current version being 0.1 returning False, so that means `major >= 1 OR minor >= 2`, I think?

tests/test_io.py
```diff
@@ -0,0 +1,96 @@
...
+    # by default re-orients the inner ring
+    result = spherely.from_wkt(polygon_with_bad_hole_wkt)
+    assert (
+        str(result)
+        == "POLYGON ((20 35, 10 30, 10 10, 30 5, 45 20, 20 35), (20 15, 20 25, 30 20, 20 15))"
+    )
+
+    # if we force to not orient, we get an error
+    with pytest.raises(RuntimeError, match="Inconsistent loop orientations detected"):
+        spherely.from_wkt(polygon_with_bad_hole_wkt, oriented=True)
```
benbovy, line 64: Out of curiosity, is there a great performance gain when setting `oriented=True` and when we know that all ring are consistently oriented? I haven't checked the implementation in s2geography (and not much in s2geometry), but I assume that the orientations are checked anyway if such case may trigger raising an error.

Maybe are there other reasons of setting `oriented=True`? I.e., when do we want to avoid implicit re-ordering of the rings?

benbovy: We might also later uniformize across the library the type of error raised (in #51 a ValueError is raised when a polygon has invalid edges).

tests/test_io.py
```diff
@@ -0,0 +1,96 @@
...
+    result = spherely.from_wkt("LINESTRING (-64 45, 0 45)")
+    assert spherely.distance(result, spherely.Point(45, -30)) > 10000
+
+    result = spherely.from_wkt("LINESTRING (-64 45, 0 45)", planar=True)
+    assert spherely.distance(result, spherely.Point(45, -30)) < 100
+
+
+@pytest.mark.skipif(
+    Version(spherely.__s2geography_version__) >= Version("0.2.0"),
+    reason="Needs s2geography >= 0.2.0",
```
benbovy, line 81:
```suggestion
    reason="Needs s2geography < 0.2.0",
```

benbovy (commented):
I added a few comments, although overall it looks already good!

I'm not sure about the cost in performance of constructing (and destroying) new `s2geography::WKTReader` (or `s2geography::WKTWriter`) objects within the vectorized inner loops, but perhaps it would be preferable to handle that out of the loop (e.g., using a Functor)? I don't think it would be very useful to allow array values for `planar` and/or `oriented`.

src/io.cpp
```diff
@@ -0,0 +1,81 @@
...
+            Set to True if polygon ring directions are known to be correct
+            (i.e., exterior rings are defined counter clockwise and interior
+            rings are defined clockwise).
+            By default (False), it will return the polygon with the smaller
+            area.
+        planar : bool, default False
+            If set to True, the edges linestrings and polygons are assumed to
+            be planar. In that case, additional points will be added to the line
+            while creating the geography objects, to ensure every point is
+            within 100m of the original line.
```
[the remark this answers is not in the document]
jorisvandenbossche: Yes we should (eg bigquery uses a default of 10m, I think), but was also thinking that this could be added later (although it is not actually difficult to add, because I already hardcode the 100m. I can maybe add this while refactoring to reuse the reader/writer object and to not vectorize those additional arguments)

src/spherely.pyi
```diff
@@ -138,6 +140,14 @@ boundary: _VFunc_Nin1_Nout1[Literal["boundary"], Geography, Geography]
 convex_hull: _VFunc_Nin1_Nout1[Literal["convex_hull"], Geography, Polygon]
 distance: _VFunc_Nin2optradius_Nout1[Literal["distance"], float, float]
 
+# io functions
+
+to_wkt: _VFunc_Nin1_Nout1[Literal["to_wkt"], str, object]
```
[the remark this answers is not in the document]
jorisvandenbossche: This `object` is the array return type, so IIUC the above is saying that the function returns a string for a single geography and an array of object dtype for an array of geographies? (which is correct I think)

tests/test_io.py
jorisvandenbossche: I don't know (without trying out) if there is a performance difference, but I think that the `oriented=True` does allow you to create different geographies? (if you have one that is larger than half of the globe, the default constructor would always normalize it to represent the smaller polygon)

And I can also imagine that just checking that the orientation of all loops are the same is still cheaper than a full normalization (but that is indeed something to check)

jorisvandenbossche: > We might also later uniformize across the library the type of error raised (in #51 a ValueError is raised when a polygon has invalid edges).

Yeah, in this case the error comes from s2geography (in theory I could catch the error and re-throw it with a different class, because those RuntimeErrors are a bit strange)

src/spherely.pyi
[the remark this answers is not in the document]
benbovy: Oops, yes you are right (I misread "to_wkt" and was thinking "from_wkt").

src/io.cpp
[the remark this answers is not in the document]
jorisvandenbossche: Added a `tessellate_tol_m` keyword, mimicking R's s2 (https://r-spatial.github.io/s2/reference/s2_geog_point.html)

src/spherely.pyi
```diff
@@ -138,6 +140,14 @@ boundary: _VFunc_Nin1_Nout1[Literal["boundary"], Geography, Geography]
...
+
+def from_wkt(
```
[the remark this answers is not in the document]
jorisvandenbossche: Used a lambda in the `m.def` so to only vectorize the main input data argument, and not oriented/planar.

jorisvandenbossche:
Updated this now https://github.com/benbovy/spherely/pull/51 is merged, I think this should be ready for a final review / merge.

src/io.cpp
```diff
@@ -0,0 +1,97 @@
...
+            By default (False), it will return the polygon with the smaller
+            area.
+        planar : bool, default False
+            If set to True, the edges of linestrings and polygons are assumed
+            to be linear on the plane. In that case, additional points will
+            be added to the line while creating the geography objects, to
+            ensure every point is within 100m of the original line.
+            By default (False), it is assumed that the edges are spherical
+            (i.e. represent the shortest path on the sphere between two points).
+        tessellate_tol_m : float, default 100.0
```
benbovy, line 78:
```suggestion
        tessellate_tol : float, default 100.0
```

(nit suggestion: `_m` makes the parameter name slightly less readable IMHO and its meaning may not be obvious at a first glance without checking the docs anyway... `_meters` would be more meaningful but would also make the parameter name too long).

benbovy (commented):
Just one minor suggestion but otherwise it is good to go!

Thanks @jorisvandenbossche !

src/io.cpp
jorisvandenbossche: That keyword name was taken from R s2 (https://r-spatial.github.io/s2/reference/s2_geog_point.html), but I am certainly fine with dropping the `_m` as well. I agree that from just seeing the keyword name (without reading the explanation for what it stands), it is not necessarily clear that this stands for "meters", making it an unnecessary suffix.

(to be honest, I am also not convinced that "tessellation" is the best name. BigQuery uses it, but if you google for it, it is typically about dividing a plane into multiple shapes that nicely cover it without overlap)

benbovy: Yes I'm more familiar with the latter meaning (dividing a surface into multiple shapes), but I'm fine with the name here.

Another name could be `densify_pts_tol` or `densify_tol` (pyproj uses [densify_pts](https://pyproj4.github.io/pyproj/dev/api/transformer.html#pyproj.transformer.Transformer.transform_bounds) for similar problems).

jorisvandenbossche: Going to merge this, so I can fixup conflicts with the geoarrow PR. But happy to change the keyword name again in a quick follow-up.

PR #51, opened 2024-10-04
Refactor Geography and add (scalar) creation functions

### LINKED ISSUE #27: Exposing various Geography subclasses?
Currently we have a Geography base class and the Point, Linestring and Polygon subclasses. `s2geography` itself has in addition also a GeographyCollection.

https://github.com/benbovy/spherely/pull/26 is adding more subclasses, and I questioned the need for that: do we want / need all those subclasses? (or at least when adding them, we should be able to answer for ourselves why we do it / what value it brings)

### LINKED ISSUE #41: Usage of lat/lng vs lng/lat in constructors
At the moment, the constructors use the lat/lng order convention from s2:

```
>>> spherely.Point(50, 4)
POINT (4 50)
```

But then, as you can see above, the WKT repr uses lng/lat, because that is (AFAIK) generally how WKT (or WKB) is expected to be expressed.

However, that is quite confusing, because that means you can't just copy the coordinates from some WKT string to the constructor ..

I was checking with R's `s2` how they do that, but I don't directly see such explicit constructors from coordinates for a geography object. They do have an explicit `s2_lnglat` to create a S2LatLng object, and indeed they also use lng/lat order there, but because of the name of the function that is also directly clear.

benbovy:
This PR refactors the Geography (sub)classes in spherely and adds geography creation scalar functions.

Supersedes #26.
Closes #41.
Closes #27.

- Goal is to remove the Geography subclasses (Point, Polygon, etc.): only one `Geography` wrapper class remains
- Geography type (and also whether the geography is empty or not) is inferred at object creation
- Geography objects are created via the top level creation functions (only scalar for now except for points)

TODO:

- [x] fix segfault when creating a geography collection itself including collection(s)
- [x] add `multipolygon` creation function
- [x] add EMPTY geographies creation function overloads (+ tests errors when trying to build MULTI geographies from EMPTY geographies)
- [x] move geography creation tests into `test_creation.py`
- [x] raise when trying to create linestring with one point only


src/accessors-geog.cpp
```diff
@@ -12,23 +14,20 @@ const double EARTH_RADIUS_METERS = 6371.01 * 1000;
...
-    return PyObjectGeography::from_geog(std::move(point));
+    return make_py_geography<s2geog::PointGeography>(s2_point);
 }
 
 PyObjectGeography boundary(PyObjectGeography a) {
     const auto& a_ptr = a.as_geog_ptr()->geog();
     auto s2_obj = s2geog::s2_boundary(a_ptr);
-    // TODO return specific subclass
     auto geog_ptr = std::make_unique<spherely::Geography>(std::move(s2_obj));
     return PyObjectGeography::from_geog(std::move(geog_ptr));
```
jorisvandenbossche, line 24: Above you made a shortcut for creating a PyObjectGeography from an s2 object (`make_py_geography`), should we do something similar for this case creating it from a s2geography object? (the name `s2_obj` here is a bit confusing, because it is a `s2geography::Geography`, not some `s2` object)

Eg add a `PyObjectGeography::from_s2geog()` to do this in one line? (this is probably not strictly related to this PR, so can also do that separately, this PR is already big)

src/creation.cpp
```diff
@@ -0,0 +1,283 @@
...
+//     ;
+// }
+
+//
+// ---- Spherely Python Geography creation functions
+//
+
+// support array_like (list of lists) of longitude, latitude coordinates
+using VectorLngLat = std::vector<std::pair<double, double>>;
+PYBIND11_MAKE_OPAQUE(VectorLngLat)
```
jorisvandenbossche, line 191: I wonder if this is needed, we could also rely on pybind11 converting a list of lists to a 2d array? (and then taking that overload)

jorisvandenbossche (commented):
I know this is not yet ready, was just already taking a quick look.

One thought regarding the scope of this PR, since it is a big one: maybe you could limit this PR to only cover what essentially already existed (to get the current tests (updated for new function names) passing). For example just provide scalar creation functions for linestring/polygon, and leave the vectorized creation functions (for everything other than point) for a follow-up.

benbovy:
Yes I've been working further on this (last updates not uploaded yet) but that might be too much of a change indeed. I'll restrict this PR to the refactoring of Geography wrapper classes.

Also I'm not exactly sure whether we should closely reuse Shapely's API for creation functions. I find that API convoluted with the mixing of a unique `coords` argument vs. separate `x`, `y` (and `z`) arguments as well as the different behavior depending on `indices`. We should probably reuse the same API for compatibility although it makes things more cumbersome to implement.

src/creation.cpp
benbovy: Ah I didn't know pybind11 supported that implicit conversion. Anyway I started (re)implementing the creation functions as it is done in shapely and we don't need this anymore.

jorisvandenbossche: To be clear I didn't test it for this case, I only know it does for our other vectorized functions.

For example used in the tests here (passing a list instead of explicit numpy array):

https://github.com/benbovy/spherely/blob/121dbf3ba8a0dd844b578de0b6ad7519930211b5/tests/test_accessors.py#L29

https://github.com/benbovy/spherely/blob/121dbf3ba8a0dd844b578de0b6ad7519930211b5/tests/test_geography.py#L139

jorisvandenbossche: > Anyway I started (re)implementing the creation functions as it is done in shapely and we don't need this anymore.

Personally I do like (at least for points) that we have the version of having both separate lon / lat arrays and a single 2d array

jorisvandenbossche:
Yeah, the vectorized creation functions in shapely are quite complex, and I am also not sure if they are always that useful (like creating polygons from a single 3D array, which I think in practice is only useful to generate some random triangles or to create bboxes), and agreed we should carefully think about which API we exactly want here. 
That's also the reason I mentioned leaving out the vectorized creation functions for now, and only providing the scalars ones (that already exist and that are used in the tests). For points, I would keep the vectorized function you already have, since in this case it is quite simple, I think? (and you already have that)

src/creation.cpp
benbovy: Ah yes I thought the conversion was maybe limited to `pybind11::vectorize` but if `spherely.create` works with lists that might be true generally.

benbovy:
> For points, I would keep the vectorized function you already have, since in this case it is quite simple, I think? (and you already have that)

Yes. For a single array this is currently limited to 2D (i.e., not a true ufunc like in shapely) but that's fine for now. I don't think that we can use `pybind11.vectorize` to extend support to a single nD input array (https://github.com/pybind/pybind11/issues/1294).

I wouldn't be too hard I think to mimic the behavior of a ufunc by relying on pybind11's array view & indexing capabilities (+ some manual computation of nD indices), though. That would be even easier to let xtensor(-python) takes care of all of that (also for creating temporary arrays of geographies on the C++ side while releasing the GIL), but I'm not sure if adding such build dependencies is worth it.

src/accessors-geog.cpp
benbovy: Yes either option is fine by me.

src/creation.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+    auto npoints = coords_data.shape(0);
+    auto points = py::array_t<PyObjectGeography>(npoints);
+
+    py::buffer_info buf = points.request();
+    py::object *data = static_cast<py::object *>(buf.ptr);
+
+    for (py::ssize_t i = 0; i < npoints; i++) {
+        auto point_ptr = point(coords_data(i, 0), coords_data(i, 1));
+        data[i] = py::cast(std::move(point_ptr));
```
jorisvandenbossche, line 126: For my education, why do we actually need the `py::cast` here? Because `point_ptr` is already a PyObjectGeography object, right?

src/accessors-geog.cpp
benbovy: I went for the latter option (`make_py_geography` overload).

src/creation.cpp
```diff
@@ -0,0 +1,436 @@
...
+        points : sequence
+            A sequence of (longitude, latitude) coordinates, in degrees.
+
+    )pbdoc");
+
+    m.def("multipoint",
+          &multipoint<Geography *>,
+          py::arg("points"),
+          R"pbdoc(
+        Create a MULTIPOINT geography from POINT geographies:
```
jorisvandenbossche, line 307: How does pybind11 handle this if you have an overloaded function (`multipoint` with list of coord pairs vs list of points), and you specify a docstring for both versions? 
For users ideally we have one docstring that mentions both options for the `points` argument?

src/creation.cpp
```diff
@@ -0,0 +1,436 @@
...
+}
+
+S2Point make_s2point(const Geography *point_ptr) {
+    check_geog_type(*point_ptr, GeographyType::Point);
+    auto s2geog_obj = static_cast<const s2geog::PointGeography &>(point_ptr->geog());
+
+    if (s2geog_obj.Points().empty()) {
+        throw EmptyGeographyException("cannot create s2geometry point from empty POINT Geography");
+        // TODO: what do we want here? NaN or S2Point default constructor? It depends?
+        // return S2Point::NaN();
```
jorisvandenbossche, line 48: Or indeed disallow as you do now? I think I would be fine not enabling the user to create a MultiPoint with empty Point

src/creation.cpp
```diff
@@ -0,0 +1,436 @@
...
+    try {
+        return make_geography<s2geog::PointGeography>(make_s2points(pts));
+    } catch (const EmptyGeographyException &error) {
+        throw py::value_error("can't create MultiPoint with empty component");
+    }
+}
+
+template <class V>
+std::unique_ptr<Geography> linestring(const std::vector<V> &pts) {
+    auto s2points = make_s2points(pts);
```
jorisvandenbossche, line 143: Also do a try/catch here like you do for multilinestring below?

src/creation.cpp
```diff
@@ -0,0 +1,436 @@
...
+        data[i] = py::cast(std::move(point_ptr));
+    }
+
+    return points;
+}
+
+template <class V>
+std::unique_ptr<Geography> multipoint(const std::vector<V> &pts) {
+    try {
+        return make_geography<s2geog::PointGeography>(make_s2points(pts));
```
jorisvandenbossche, line 135: Those functions don't have to return a PyObjectGeography (like `point` above does) to do the automatic binding for the python function equivalent? (i.e. using `make_py_geography` instead of `make_geography`)

jorisvandenbossche:
Nice progress! 
One thought on the API: given that we don't follow exact shapely anymore with the subclass constructors (`shapely.Point(..)` etc), we _could_ also follow the naming of s2 and bigquery and others for the scalar creation functions, and use something like `make_point()` instead of `point()`, etc. 
(although I see they are also not consistent in it, as they have `st_geogpoint` and `st_makeline`/`st_makepolygon` ...)

(I am not necessarily convinced myself that this is better (it's longer to type, but tab completion will put them all together, and there is a clearer distinction from being a type), mostly just mentioning it in case you would have a stronger opinion)

src/creation.cpp
benbovy: Pybind11 concatenates the docstrings and enumerate the signature of each overloaded function.

Since the arguments don't change but their types does, I disabled the auto-generated signatures and manually added a custom, generic one in the docstrings of the 1st overload (you can have a look at the rendered API documentation for this PR on RTD).

Alternatively we could define docstrings only for the 1st overload such that there is one "Parameters" section with mixed types.

src/creation.cpp
benbovy: We don't have to do it here (we expose `Geography` as a `py::class_` so pybind11 does the automatic binding).

PyObjectGeography is needed when using geographies with `py::array` or `py::vectorize` since pybind11 does not support the object dtype out of the box (actually this is probably  supported it now but not sure it works without copy).

src/creation.cpp
benbovy: Yes raising an error is good in the context here. This comment was in case we later reuse this function in other contexts (I'll clarify it).

src/creation.cpp
benbovy: We don't need it, actually :-) (lines of code naively copied from old snippets).

benbovy:
Thanks for the review comments @jorisvandenbossche.

I also have been thinking about the naming... Likewise I'm still not fully convinced myself and I don't have strong opinions either.

We could also use the `create_` prefix for what we call "geography creation" functions...

src/creation.cpp
jorisvandenbossche: Ah, thanks for that explanation! 
That also explains why the functions in `accessors-geog.cpp` need to return PyObjectGeography explicitly, since those are used with `vectorized`

src/creation.cpp
jorisvandenbossche: Seeing the example in https://spherely--51.org.readthedocs.build/en/51/_api_generated/spherely.multipoint.html, I think I would have a preference for manually writing a combined parameters section. But that is also something that can be addressed later!

tests/test_accessors.py
```diff
@@ -7,39 +7,39 @@
...
+            spherely.polygon([(0, 0), (0, 2), (2, 2), (2, 0)]),
+            spherely.point(1, 1),
         ),
     ],
 )
 def test_centroid(geog, expected) -> None:
     # scalar
     actual = spherely.centroid(geog)
-    assert isinstance(actual, spherely.Point)
+    assert spherely.get_type_id(actual) == spherely.GeographyType.POINT.value
```
jorisvandenbossche, line 24:
```suggestion
    assert spherely.get_type_id(actual) == spherely.GeographyType.POINT
```

It doesn't work without the `.value`? 

I would expect that an int-based enum can be compared to integers directly

tests/test_geography.py
```diff
@@ -5,59 +5,117 @@
...
+
+    with pytest.raises(ValueError, match="polygon is not valid.*crosses.*"):
+        # shell/hole rings are crossing each other
+        spherely.polygon(
+            shell=[(0, 0), (0, 4), (4, 4), (4, 0)],
+            holes=[[(0, 1), (0, 5), (5, 5), (5, 1)]],
+        )
 
 
 def test_polygon_normalize() -> None:
```
jorisvandenbossche, line 114: Should probably also add a test here with holes, to verify those also get normalized/oriented?

tests/test_geography.py
```diff
@@ -68,32 +126,79 @@ def test_polygon_normalize() -> None:
...
+    assert repr(coll).startswith("GEOMETRYCOLLECTION (")
+
+    # TODO: more robust test than using the WKT repr
+    assert repr(coll).count("POINT") == 1
+    assert repr(coll).count("LINESTRING") == 1
+    assert repr(coll).count("POLYGON") == 1
+
+    # TODO: test objects are copied
+    # (for now only test that original objects are preserved)
+    assert [o.nshape for o in objs] == [1, 1, 1]
```
jorisvandenbossche, line 149: Test `coll.nshape` here as well?

tests/test_geography.py
```diff
@@ -68,32 +126,79 @@ def test_polygon_normalize() -> None:
...
+
+    # test nested collection
+    # coll2 = spherely.geography_collection(objs + [coll])
```
jorisvandenbossche, line 152: This is not yet supported? (does that require custom handling? FWIW I would be fine with not supporting this)

tests/test_predicates.py
```diff
@@ -97,6 +97,19 @@ def test_disjoint():
...
 
     # two scalars
-    a2 = spherely.Point(50, 9)
-    b2 = spherely.LineString([(50, 8), (60, 8)])
+    a2 = spherely.point(50, 9)
+    b2 = spherely.linestring([(50, 8), (60, 8)])
     assert spherely.disjoint(a2, b2)
+
+
+def test_predicates_polygon():
```
jorisvandenbossche, line 105: Move above to below `test_contains` as `test_contains_polygon` ? (given this is mostly testing `contains`)

(I suppose the reason those tests are using points/linestrings is because polygon creation was not yet added when writing those tests?)

jorisvandenbossche (commented):
A few more comments!

tests/test_geography.py
benbovy: It used to work but not anymore after refactoring Geography (I get a segfault locally).

tests/test_geography.py
benbovy: There's already a `assert coll.nshape == 3` a few lines above, unless you mean something else?

src/creation.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+namespace spherely {
+
+//
+// ---- S2geometry / S2Geography / Spherely object wrapper utility functions
+//
+
+/*
+** Wrap one or more s2geometry objects into a spherely::Geography object.
+**
```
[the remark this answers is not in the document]
benbovy: Done.

src/creation.cpp
benbovy: That wasn't much work to address so I did it here.

tests/test_geography.py
benbovy: Fixed.

tests/test_accessors.py
jorisvandenbossche: Hmm, that's unfortunate ..

src/geography.cpp
```diff
@@ -26,128 +30,137 @@ using namespace spherely;
...
+        const auto &s2poly_ptr = ptr->Polygon();
+        // find the outer shells (loop depth = 0, 2, 4, etc.)
+        std::vector<int> outer_shell_loop_ids;
+
+        outer_shell_loop_ids.reserve(static_cast<size_t>(s2poly_ptr->num_loops()));
+
+        for (int i = 0; i < s2poly_ptr->num_loops(); i++) {
+            if ((s2poly_ptr->loop(i)->depth() % 2) == 0) {
+                outer_shell_loop_ids.push_back(i);
+            }
```
jorisvandenbossche, line 145: Given that you don't really use the `outer_shell_loop_ids` vector, except for checking the size, you could also replace this with a simpler count?

```
int n_outer_shell_loops = 0;
...
for .. 
  if .. {
    n_outer_shell_loops ++;
  }
```

tests/test_creation.py
```diff
@@ -0,0 +1,315 @@
...
+def test_collection() -> None:
+    objs = [
+        spherely.point(0, 0),
+        spherely.linestring([(0, 0), (1, 1)]),
+        spherely.polygon([(0, 0), (1, 0), (1, 1)]),
+    ]
+
+    coll = spherely.geography_collection(objs)
+
+    assert coll.dimensions == -1
```
jorisvandenbossche, line 296: This doesn't match with the docstring, stating that for a collection it is the max dimension of the individual features

jorisvandenbossche (approved):
Nice!

jorisvandenbossche:
> This is getting close.
> 
> A few other comments that we can address later:

Agreed! Let's indeed get this in (to have the basics working, and will also be easier for other PRs adding features), and the things below can be handled in follow-ups.

> * Empty geographies: there is no overload function to create MULTI geographies without argument (or with None) ; also when empty or 1-length sequences are given the multi-types are demoted (see [Singleton Multi-types automatically demoted paleolimbot/s2geography#17](https://github.com/paleolimbot/s2geography/issues/17))

Personally I am fine with that limitation for now, especially given that `s2geography` essentially does not make the distinction between single and multi geometries anyway (as you linked), and that it is only for output (like WKT, or asking for the type) that the distinction is introduced, based on the number of elements stored in the object.

> * The collection type is named `GeographyType::GeographyCollection` but both the Geography repr and WKT have the  `GEOMETRYCOLLECTION` name. Do we want to uniformize things and rename `GeographyType::GeographyCollection` to `GeographyType::GeometryCollection`?

I would maybe lean towards `GeographyType::GeometryCollection`, although that is certainly odd as well .. 

Another case where this occurs is in the `spherely.geography_collection(..)` constructor. For that case, we could also shorten it to just `collection()` to avoid that issue? (or at least when we add a prefix like `create_..`, that might be sufficient)
 
> * It is not currently possible to create a MULTIPOLYGON with duplicate rings, due to the fact that `s2geography::PolygonGeography` reuse the same underlying `S2Polygon` for multi-polygons. Maybe if we deactivate s2 normalization and/or validity check it will pass but will it work ?

A multipolygon with duplicate rings also sounds like "wrong" .. (in shapely / GEOS, that would be an invalid geometry anyway, and thus give unexpected results), so I am personally fine with not allowing to create it. 

Although I suppose at some point we should allow to create invalid geographies, end then have a method to check if they are valid / what the reason is for the invalidity (for example [`s2_make_polygon`](https://r-spatial.github.io/s2/reference/s2_geog_point.html) has a `check` keyword to disable validation when constructing a polygon)

This is also another case (together with the geometrycollection / geographycollection above) where it points the the mismatch between the "simple features" model we try to emulate and how s2 works. 
For the constructor functions you implemented here, it is nice to make the distinction between all the types, I think (as it gives a more user friendly interface), but for accessors like the geography type, we could also stay more closely to s2geography and have a function that just returns point/line/polygon/collection

> * Scalar creation function naming: still no strong opinion, although I would lean towards either `point()`, `linestring()`, etc. or `create_point()`, `create_linestring()`, etc.

Let's open a separate issue for this and see if we can get some more opinions

src/geography.hpp
```diff
@@ -14,116 +20,148 @@ using S2GeographyIndexPtr = std::unique_ptr<s2geog::ShapeIndexGeography>;
...
+** - Encapsulate a lazy ``s2geography::ShapeIndexGeography`` accessible via ``geog_index()``.
 **
 */
 class Geography {
 public:
     Geography(const Geography&) = delete;
     Geography(Geography&& geog) : m_s2geog_ptr(std::move(geog.m_s2geog_ptr)) {
-        // std::cout << "Geography move constructor called: " << this <<
-        // std::endl;
+        extract_geog_properties();
```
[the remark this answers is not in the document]
benbovy: > Actually, I should have commented that on the case below (creating from s2geography), because for the line above in theory we could propagate the attributes from the passed Geography object?

Yes indeed we should simply propagate the attributes in this case.

In the case of creating from an s2geography object when we know the geography type, we could indeed provide the type explicitly instead of inferring it (and therefore avoid the inference overhead). We certainly want to avoid passing inconsistent s2geography object type vs. geography type, though, in order to prevent odd behavior or segfaults happening later. Checking for this consistency at runtime would be pretty similar to `extract_geog_properties` in terms of performance overhead I think. If we really want to optimize this, perhaps we could use template constructors + SFINAE or `if constexpr`...

tests/test_creation.py
benbovy: Opened #57.

benbovy:
> I would maybe lean towards GeographyType::GeometryCollection, although that is certainly odd as well ..

> Another case where this occurs is in the spherely.geography_collection(..) constructor. For that case, we could also shorten it to just collection() to avoid that issue?

I renamed `GeographyType::GeographyCollection` -> `GeographyType::GeometryCollection` and `geography_collection()` to `collection()`, I also think it looks nicer for now.

jorisvandenbossche (approved):

jorisvandenbossche:
OK, merged this, so we can move forward with other PRs and follow-ups. Thanks @benbovy!

PR #53, opened 2024-10-09
ENH: export to geoarrow output

jorisvandenbossche:
Adding geoarrow export (after import was added in https://github.com/benbovy/spherely/pull/49)

This depends on https://github.com/paleolimbot/s2geography/pull/30 (merged now, but not yet released) and https://github.com/paleolimbot/s2geography/pull/37 (for tesseltation/projection)


src/geoarrow.cpp
```diff
@@ -71,7 +71,178 @@ py::array_t<PyObjectGeography> from_geoarrow(py::object input,
...
+    }
+
+    ~ArrowArrayHolder() {
+        reset();
+    }
+
+protected:
+    ArrowArray array_;
+    ArrowSchema schema_;
+};
```
jorisvandenbossche, line 196: @paleolimbot could you take a quick look at this `ArrowArrayHolder` class if that seems to make sense? 
It's mostly inspired on the UniqueArray/Schema from nanoarrow.hpp (but then combined in a single class, making it easier to expose in python), but since I only need this (for now), it didn't feel worth vendoring the entire of nanoarrow for the 120 lines of code above.

src/geoarrow.cpp
```diff
@@ -71,7 +71,178 @@ py::array_t<PyObjectGeography> from_geoarrow(py::object input,
...
+    s2geog::geoarrow::Writer writer;
+    std::vector<std::unique_ptr<s2geog::Geography>> s2geog_vec;
+
+    s2geog::geoarrow::ExportOptions options;
+    options.set_precision(precision);
+    if (planar) {
+        auto tol = S1Angle::Radians(tessellate_tolerance / EARTH_RADIUS_METERS);
+        options.set_tessellate_tolerance(tol);
+        // TODO make this configurable
+        // options.set_projection(s2geog::geoarrow::mercator());
```
jorisvandenbossche, line 215: @benbovy we want to make this configurable, so the user can specify one of the pre-defined projections in s2geography (see https://github.com/paleolimbot/s2geography/pull/39 for the options in s2geograhy), but any thoughts on the user API to do this?

We could have a very shallow `Projection` class that essentially just holds a S2::Projection, and then some creation functions for the different projections. 

For example something with class methods like `spherely.Projection.pseudo_mercator()` ? Or just functions like `spherely.pseudo_mercator_projection()`? (or `spherely.projection_pseudo_mercator()`)

jorisvandenbossche: (and R `s2` has `s2_projection_plate_carree()`, `s2_projection_mercator()`, `s2_projection_orthographic(centre=(0,0))` , listed on https://r-spatial.github.io/s2/reference/wk_handle.s2_geography.html)

src/geoarrow.cpp
paleolimbot: I don't see any issues here!

It's also possible to use a `std::unique_ptr<>` if you don't feel like maintaining this:

https://github.com/rapidsai/cudf/blob/branch-24.12/cpp/include/cudf/interop.hpp#L121

https://github.com/rapidsai/cudf/blob/branch-24.12/cpp/src/interop/to_arrow_schema.cpp#L227-L230

...then what you have here basically collapses to `std::pair<unique_schema_t, unique_array_t>`.

jorisvandenbossche: Interesting. Now, I still need to implement the "move" logic for the method to create and return the capsules (and the `return_capsules` method itself), and that together is already more than half of the code here. So not sure if using `std::unique_ptr` would end up that much simpler (and I also need a class to expose that as an object in python)

src/geoarrow.cpp
benbovy: I like the shallow `Projection` class with class methods.

jorisvandenbossche: I have a follow-up PR exposing projections using a shallow `spherely.Projection` class -> https://github.com/benbovy/spherely/pull/69

src/geoarrow.cpp
```diff
@@ -71,7 +71,178 @@ py::array_t<PyObjectGeography> from_geoarrow(py::object input,
...
+    }
+
+protected:
+    ArrowArray array_;
+    ArrowSchema schema_;
+};
+
+ArrowArrayHolder to_geoarrow(py::array_t<PyObjectGeography> input,
+                             py::object output_schema,
+                             //  py::object geometry_encoding,
```
benbovy, line 200: Is it left for a follow-up PR?

src/geoarrow.cpp
```diff
@@ -71,7 +71,178 @@ py::array_t<PyObjectGeography> from_geoarrow(py::object input,
...
+    options.set_precision(precision);
+    if (planar) {
+        auto tol = S1Angle::Radians(tessellate_tolerance / EARTH_RADIUS_METERS);
+        options.set_tessellate_tolerance(tol);
+        // TODO make this configurable
+        // options.set_projection(s2geog::geoarrow::mercator());
+    }
+
+    if (!output_schema.is(py::none())) {
+        py::capsule schema_capsule = output_schema.attr("__arrow_c_schema__")();
```
benbovy, line 219: I guess if `output_schema` is neither None nor has `__arrow_c_schema__` defined this will raise an AttributeError here? Is that fine or should we raise a more informative error in that case?

src/geoarrow.cpp
```diff
@@ -71,7 +71,178 @@ py::array_t<PyObjectGeography> from_geoarrow(py::object input,
...
+        ArrowSchema* schema = static_cast<ArrowSchema*>(schema_capsule);
+        writer.Init(schema, options);
+        array.set_schema(schema);
+        // } else if (geometry_encoding.equal(py::str("WKT"))) {
+        //     writer.Init(s2geog::geoarrow::Writer::OutputType::kWKT, options);
+        // } else if (geometry_encoding.equal(py::str("WKB"))) {
+        //     writer.Init(s2geog::geoarrow::Writer::OutputType::kWKB, options);
```
benbovy, line 226: Remove? Or comment if that is for a follow-up PR?

src/geoarrow.cpp
```diff
@@ -71,7 +71,178 @@ py::array_t<PyObjectGeography> from_geoarrow(py::object input,
...
+    } else {
+        throw std::invalid_argument(
+            "'output_schema' should be specified or 'geometry_encoding' should be one of None, "
+            "'WKT' or 'WKB'");
```
benbovy, line 230: Missing a unit test for this case, unless I overlooked it.

src/geoarrow.cpp
```diff
@@ -124,4 +295,68 @@ void init_geoarrow(py::module& m) {
...
             respectively.
     )pbdoc");
+
+    m.def("to_geoarrow",
+          &to_geoarrow,
+          py::arg("input"),
+          py::pos_only(),
+          py::kw_only(),
+          py::arg("output_schema") = py::none(),
+          //   py::arg("geometry_encoding") = py::none(),
```
benbovy, line 305: Remove? (cf. comments above)

src/geoarrow.cpp
```diff
@@ -124,4 +295,68 @@ void init_geoarrow(py::module& m) {
...
+            i.e. which has a ``__arrow_c_schema__`` method). For example, this
+            can be a ``pyarrow.DataType`` or ``pyarrow.Field``, and you can
+            use the ``geoarrow.pyarrow`` package to construct such geoarrow
+            extension types.
+        planar : bool, default False
+            If set to True, the edges of linestrings and polygons in the output
+            are assumed to be linear on the plane. In that case, additional
+            points will be added to the line while converting to the output
+            encoding, to ensure every point is within 100m of the original
+            line on the sphere.
```
benbovy, line 334:
```suggestion
            encoding, to ensure every point is within the ``tessellate_tolerance``
            distance of the original line on the sphere.
```

tests/test_geoarrow.py
```diff
@@ -109,3 +109,55 @@ def test_from_geoarrow_invalid_encoding():
...
+
+
+def test_to_geoarrow_wkb():
+    arr = spherely.points([1, 2, 3], [1, 2, 3])
+    result = pa.array(spherely.to_geoarrow(arr, output_schema=ga.wkb()))
+    expected = ga.as_wkb(
+        [
+            "POINT (0.9999999999999998 1)",
+            "POINT (2 1.9999999999999996)",
+            "POINT (3.0000000000000004 3.0000000000000004)",
```
benbovy, line 142: This looks a bit weird?

benbovy (commented):
Just left a few minor comments.

src/geoarrow.cpp
jorisvandenbossche: Yeah, just as for the input function, for WKB/WKT I would like that the user could just specify that without having the create a matching arrow schema, for convenience. But s2geography does not yet expose the necessary tools to do so (it already allow to specify WKT/WKB as output type, but then we would still have to create a ArrowSchema struct to create a valid output `ArrowArrayHolder`)

jorisvandenbossche: But I can also leave this commented part out to clean things up

src/geoarrow.cpp
jorisvandenbossche: I added a more informative error message in the from_geoarrow code, so can do that here as well

src/geoarrow.cpp
jorisvandenbossche: Yes, see my answer above, I would like to enable this, but we still lack the easy way to create the corresponding ArrowSchema if the user did not provide this. 

Will add a comment

tests/test_geoarrow.py
jorisvandenbossche: Yes, this looks weird, but I don't know a better way to test it ..

The general problem with those test assertions is that we test in lon/lat space, but s2geometry converts the input lon/lat to unit vector, and then converting back to lon/lat (to WKB in this case) will not give the exact same values as the input values. 

For testing with WKT, this is generally fine because we limit the precision of the output, and then you don't see those small differences. But for the WKB output, as a binary value, that contains the full exact float values, so the above is needed if we want to verify the binary array result.

Of course, I could also then convert the WKB again to WKT and just verify that, but that is an additional step while I wanted to test the WKB result more directly (I am of course still creating the expected WKB from WKT, so maybe I could then just as well convert the result WKB to WKT and just verify that with the expected WKT ...)

benbovy: Thanks for the explanation! To me that's fine leaving it like this (maybe with a small comment summarizing what you explain here above).

jorisvandenbossche: OK, added a comment!

benbovy (approved):
Thank you Joris!

PR #56, opened 2024-10-20
ENH: Added more missing predicates

JoelJaeschke:
This PR adds the remaining predicates from #17.

I would greatly appreciate your feedback @benbovy regarding `covers` and `covered_by`, as the way it is implemented now appears to mostly work, but fails when checking polygons that partly share a boundary. This is consistent with https://github.com/r-spatial/s2/blob/b495b0df53bffc7dc1ad3780fe8ac208e78cd1bf/R/s2-predicates.R#L121C1-L124C2 for example, but deviates from PostGIS IIUC (cmp. https://postgis.net/docs/ST_Covers.html). From a user perspective, I would expect the test to work.

### DIFF
--- a/src/predicates.cpp
+++ b/src/predicates.cpp
@@ -117,4 +117,77 @@ void init_predicates(py::module& m) {
             Geography object(s)
 
     )pbdoc");
+
+    m.def("touches",
+          py::vectorize(Predicate([](const s2geog::ShapeIndexGeography& a_index,
+                                     const s2geog::ShapeIndexGeography& b_index,
+                                     const S2BooleanOperation::Options& options) {
+              S2BooleanOperation::Options closedOptions = options;
+              closedOptions.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
+              closedOptions.set_polyline_model(S2BooleanOperation::PolylineModel::CLOSED);
+
+              S2BooleanOperation::Options openOptions = options;
+              openOptions.set_polygon_model(S2BooleanOperation::PolygonModel::OPEN);
+              openOptions.set_polyline_model(S2BooleanOperation::PolylineModel::OPEN);
+
+              return s2geog::s2_intersects(a_index, b_index, closedOptions) &&
+                     !s2geog::s2_intersects(a_index, b_index, openOptions);
+          })),
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if A and B intersect, but their interiors do not intersect.
+        A and B have at least one point in common, where the common point
+        lies in at least one boundary.
+
+        Parameters
+        ----------
+        a, b : :py:class:`Geography` or array_like
+            Geography object(s)
+
+    )pbdoc");
+
+    m.def("covers",
+          py::vectorize(Predicate([](const s2geog::ShapeIndexGeography& a_index,
+                                     const s2geog::ShapeIndexGeography& b_index,
+                                     const S2BooleanOperation::Options& options) {
+              S2BooleanOperation::Options closedOptions = options;
+              closedOptions.set_polyline_model(S2BooleanOperation::PolylineModel::CLOSED);
+              closedOptions.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
+
+              return s2geog::s2_contains(a_index, b_index, closedOptions);
+          })),
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if every point in B lies inside the interior or boundary of A.
+
+        Parameters
+        ----------
+        a, b : :py:class:`Geography` or array_like
+            Geography object(s)
+
+    )pbdoc");
+
+    m.def("covered_by",
+          py::vectorize(Predicate([](const s2geog::ShapeIndexGeography& a_index,
+                                     const s2geog::ShapeIndexGeography& b_index,
+                                     const S2BooleanOperation::Options& options) {
+              S2BooleanOperation::Options closedOptions = options;
+              closedOptions.set_polyline_model(S2BooleanOperation::PolylineModel::CLOSED);
+              closedOptions.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
+
+              return s2geog::s2_contains(b_index, a_index, closedOptions);
+          })),
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if every point in A lies inside the interior or boundary of B.
+
+        Parameters
+        ----------
+        a, b : :py:class:`Geography` or array_like
+            Geography object(s)
+
+    )pbdoc");
 }
--- a/src/predicates.cpp
+++ b/src/predicates.cpp
@@ -155,7 +155,7 @@ void init_predicates(py::module& m) {
           py::arg("b"),
           R"pbdoc(
         Returns True if A and B intersect, but their interiors do not intersect.
-        
+
         A and B must have at least one point in common, where the common point
         lies in at least one boundary.
 
@@ -193,7 +193,7 @@ void init_predicates(py::module& m) {
         If A and B are both polygons and share co-linear edges,
         `covers` currently returns expected results only when those
         shared edges are identical.
- 
+
     )pbdoc");
 
     m.def("covered_by",
--- a/src/predicates.cpp
+++ b/src/predicates.cpp
@@ -155,7 +155,8 @@ void init_predicates(py::module& m) {
           py::arg("b"),
           R"pbdoc(
         Returns True if A and B intersect, but their interiors do not intersect.
-        A and B have at least one point in common, where the common point
+        
+        A and B must have at least one point in common, where the common point
         lies in at least one boundary.
 
         Parameters
@@ -187,6 +188,12 @@ void init_predicates(py::module& m) {
         a, b : :py:class:`Geography` or array_like
             Geography object(s)
 
+        Notes
+        -----
+        If A and B are both polygons and share co-linear edges,
+        `covers` currently returns expected results only when those
+        shared edges are identical.
+ 
     )pbdoc");
 
     m.def("covered_by",
@@ -207,5 +214,8 @@ void init_predicates(py::module& m) {
         a, b : :py:class:`Geography` or array_like
             Geography object(s)
 
+        See Also
+        --------
+        covers
     )pbdoc");
 }
--- a/src/predicates.cpp
+++ b/src/predicates.cpp
@@ -22,6 +22,10 @@ public:
     template <class F>
     Predicate(F&& func) : m_func(std::forward<F>(func)) {}
 
+    template <class F>
+    Predicate(F&& func, const S2BooleanOperation::Options& options)
+        : m_func(std::forward<F>(func)), m_options(options) {}
+
     bool operator()(PyObjectGeography a, PyObjectGeography b) const {
         const auto& a_index = a.as_geog_ptr()->geog_index();
         const auto& b_index = b.as_geog_ptr()->geog_index();
@@ -31,6 +35,33 @@ public:
 private:
     FuncType m_func;
     S2BooleanOperation::Options m_options;
+};
+
+/*
+** A specialization of the `Predicate` class for the touches operation,
+** as two `S2BooleanOpteration::Options` objects are necessary.
+*/
+class TouchesPredicate {
+public:
+    TouchesPredicate() : m_closed_options(), m_open_options() {
+        m_closed_options.set_polyline_model(S2BooleanOperation::PolylineModel::CLOSED);
+        m_closed_options.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
+
+        m_open_options.set_polyline_model(S2BooleanOperation::PolylineModel::OPEN);
+        m_open_options.set_polygon_model(S2BooleanOperation::PolygonModel::OPEN);
+    }
+
+    bool operator()(PyObjectGeography a, PyObjectGeography b) const {
+        const auto& a_index = a.as_geog_ptr()->geog_index();
+        const auto& b_index = b.as_geog_ptr()->geog_index();
+
+        return s2geog::s2_intersects(a_index, b_index, m_closed_options) &&
+               !s2geog::s2_intersects(a_index, b_index, m_open_options);
+    }
+
+private:
+    S2BooleanOperation::Options m_closed_options;
+    S2BooleanOperation::Options m_open_options;
 };
 
 void init_predicates(py::module& m) {
@@ -119,20 +150,7 @@ void init_predicates(py::module& m) {
     )pbdoc");
 
     m.def("touches",
-          py::vectorize(Predicate([](const s2geog::ShapeIndexGeography& a_index,
-                                     const s2geog::ShapeIndexGeography& b_index,
-                                     const S2BooleanOperation::Options& options) {
-              S2BooleanOperation::Options closedOptions = options;
-              closedOptions.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
-              closedOptions.set_polyline_model(S2BooleanOperation::PolylineModel::CLOSED);
-
-              S2BooleanOperation::Options openOptions = options;
-              openOptions.set_polygon_model(S2BooleanOperation::PolygonModel::OPEN);
-              openOptions.set_polyline_model(S2BooleanOperation::PolylineModel::OPEN);
-
-              return s2geog::s2_intersects(a_index, b_index, closedOptions) &&
-                     !s2geog::s2_intersects(a_index, b_index, openOptions);
-          })),
+          py::vectorize(TouchesPredicate()),
           py::arg("a"),
           py::arg("b"),
           R"pbdoc(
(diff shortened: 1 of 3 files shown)


benbovy:
Thanks for the PR @JoelJaeschke !

I suspect that that R s2 deviating from PostGIS is not intentional? If we want a behavior consistent with PostGIS the implementation of `covers` should probably check both "s2_contains" and "s2_intersects" (not only "s2_contains").

I briefly checked in the S2geometry tests: for two polygons sharing one edge, the open vs. closed polygon model has no influence on the result of the DIFFERENCE Boolean operation (used by "s2_contains") but yields different results for the INTERSECTION operation.

(S2geometry also seems to assume that the edges of polylines - and polygons? - are directed by default and takes that into account for the intersection operation... this might also explain some specific behavior).

JoelJaeschke:
Actually, what I thought was incorrect behavior is likely expected and working correctly. I was thinking of this from a planar geometry point of view, but since S2 assumes a spherical earth, the horizontal edges are not actually straight, but rather follow the earth's curvature. Therefore, even though in planar geometry, one polygon would be covered, in spherical geometry this is not true anymore.

I tested this in BigQuery (using the GeoViz tool) and the following query
```sql
select st_geogfromtext('polygon((-118 60, -118 23, 34 23, 34 60, -118 60))') p1, st_covers(st_geogfromtext('polygon((-118 60, -118 23, 34 23, 34 60, -118 60))'), st_geogfromtext('polygon((-118 60, -118 23, 34 23, 34 60, -118 60))'))
union all
select st_geogfromtext('polygon((-118 60, -118 23, -18 23, -18 60, -118 60))') p2, st_covers(st_geogfromtext('polygon((-118 60, -118 23, 34 23, 34 60, -118 60))'), st_geogfromtext('polygon((-118 60, -118 23, -18 23, -18 60, -118 60))'));
```
where in the second select, the polygon is similar to the way I wrote the tests. However, when visualizing the resulting geometries, this is the result:
Screenshot from 2024-10-21 19-14-30
As can be seen, the red polygon is definitely not covered by the "parent" polygon. BigQuery also returns false for the second `ST_Covers`. So I think the test rightfully fails. Let me know if this makes sense to you @benbovy . Otherwise, I would simply remove the last test from the PR and it should be good to go.

benbovy:
Ah yes that makes sense!

(BTW we should implement quick visualization of geography objects in Spherely as it helps a lot!)

Instead of simply removing the failing tests, could you please keep these and make sure that the boundaries of the two polygons partially overlap along a great circle (longitude)?

(I'm also curious to see if it works as expected when the overlapping boundaries correspond to a shell ring for polygon A and a hole ring for polygon B... I guess it should work?)

JoelJaeschke:
> Instead of simply removing the failing tests, could you please keep these and make sure that the boundaries of the two polygons partially overlap along a great circle (longitude)?

This has turned out to be much tougher than I anticipated. The standard `S2BooleanOperation::Contains` in s2geometry does not handle shared boundaries on polygons as I had hoped.. (well, it does, but not in a way that would help when implementing the covers logic, see https://github.com/google/s2geometry/issues/387).

I am not sure whether I grok the concepts in s2geometry fully, but as I understood, there is no API that handles the covers check natively. The `S2Polygon` class has a method called `ApproxContains`, which solves the issue of shared boundaries on polygons, but only when using an approximation, i.e. no exact checks. However, this would require some custom dispatching logic that differs between polygons and polylines/points and I am not quite sure whether this is something that should be implemented in spherely or rather in s2geography?

Below is some code that demonstrates the issue in more detail (only depends on s2geometry):
```cpp
#include <iostream>
#include <memory>

#include <s2/s2latlng.h>
#include <s2/s2point.h>
#include <s2/s2loop.h>
#include <s2/s2polygon.h>
#include <s2/s2builder.h>
#include <s2/s2builderutil_s2polygon_layer.h>
#include <s2/s2boolean_operation.h>

int main(int argc, char** argv) {
    const std::vector<S2Point> parent_vertices = {
        S2LatLng::FromDegrees(60, -118).ToPoint(),
        S2LatLng::FromDegrees(23, -118).ToPoint(),
        S2LatLng::FromDegrees(23, 34).ToPoint(),
        S2LatLng::FromDegrees(60, 34).ToPoint()
    };
    std::unique_ptr<S2Loop> parent_loop = std::make_unique<S2Loop>(std::move(parent_vertices));
    S2Polygon parent_poly(std::move(parent_loop));

    const std::vector<S2Point> interior_vertices = {
        S2LatLng::FromDegrees(40, -117).ToPoint(),
        S2LatLng::FromDegrees(37, -117).ToPoint(),
        S2LatLng::FromDegrees(37, -116).ToPoint(),
        S2LatLng::FromDegrees(40, -116).ToPoint()
    };
    std::unique_ptr<S2Loop> interior_loop = std::make_unique<S2Loop>(std::move(interior_vertices));
    S2Polygon interior_poly(std::move(interior_loop));

    const std::vector<S2Point> shared_bound_vertices = {
        S2LatLng::FromDegrees(40, -118).ToPoint(),
        S2LatLng::FromDegrees(23, -118).ToPoint(),
        S2LatLng::FromDegrees(23, 34).ToPoint(),
        S2LatLng::FromDegrees(40, 34).ToPoint()
    };
    std::unique_ptr<S2Loop> shared_bound_loop = std::make_unique<S2Loop>(std::move(shared_bound_vertices));
    S2Polygon shared_bound_poly(std::move(shared_bound_loop));
    
    const std::vector<S2Point> crossing_vertices = {
        S2LatLng::FromDegrees(40, -120).ToPoint(),
        S2LatLng::FromDegrees(37, -120).ToPoint(),
        S2LatLng::FromDegrees(37, -116).ToPoint(),
        S2LatLng::FromDegrees(40, -116).ToPoint()
    };
    std::unique_ptr<S2Loop> crossing_loop = std::make_unique<S2Loop>(std::move(crossing_vertices));
    S2Polygon crossing_poly(std::move(crossing_loop));
   
    S2BooleanOperation::Options options;
    options.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
    options.set_polyline_model(S2BooleanOperation::PolylineModel::OPEN);

    const auto contains_interior = S2BooleanOperation::Contains(parent_poly.index(), interior_poly.index()) ? "true" : "false";
    std::cout << "Contains interior? " << contains_interior << std::endl;

    const auto contains_shared = S2BooleanOperation::Contains(parent_poly.index(), shared_bound_poly.index()) ? "true" : "false";
    std::cout << "Contains shared? " << contains_shared << std::endl;

    const S1Angle tol = S1Angle::Degrees(1e-16);
    const auto approx_contains_shared = parent_poly.ApproxContains(shared_bound_poly, tol) ? "true" : "false";
    std::cout << "Approx contains shared? " << approx_contains_shared << std::endl;

    const auto contains_crossing = S2BooleanOperation::Contains(parent_poly.index(), crossing_poly.index()) ? "true" : "false";
    std::cout << "Contains crossing? " << contains_crossing << std::endl;

    const auto approx_contains_crossing = parent_poly.ApproxContains(crossing_poly, tol) ? "true" : "false";
    std::cout << "Approx contains crossing? " << approx_contains_crossing << std::endl;

    return 0;
}
```
Output when called is
```
Contains interior? true
Contains shared? false
Approx contains shared? true
Contains crossing? false
Approx contains crossing? false
```

jorisvandenbossche:
@JoelJaeschke thanks for further diving into this!

Exploring this using R's `s2` (given this is already implemented there), this is a reproducer of your example above I think (with the shared boundary that gives a wrong result):

```r
# s2_make_polygon takes vector of longitudes and vector of latitudes
> parent_poly <- s2_make_polygon(c(-118, -118, 34, 34), c(60, 23, 23, 60))
> shared_bound_poly <- s2_make_polygon(c(-118, -118, 34, 34), c(40, 23, 23, 40))
> s2_contains(parent_poly, shared_bound_poly)
[1] FALSE
> s2_covers(parent_poly, shared_bound_poly)
[1] FALSE
```

and this indeed gives False while you would expect True for both cases (so this is actually not specific to `covers` being added here, but already an issue with `contains` as well, AFAIU)

Testing what has been said on the s2geometry, that this is a 50/50 chance of falling left or right of the edge, I tested with some other coordinates along the edge:
```r
> for (lat_high in 40:30) {
+     shared_bound_poly <- s2_make_polygon(c(-118, -118, 34, 34), c(lat_high, 23, 23, lat_high))
+     print(paste(lat_high, s2_contains(parent_poly, shared_bound_poly), s2_covers(parent_poly, shared_bound_poly)))
+ }
[1] "40 FALSE FALSE"
[1] "39 FALSE FALSE"
[1] "38 TRUE TRUE"
[1] "37 FALSE FALSE"
[1] "36 FALSE FALSE"
[1] "35 TRUE TRUE"
[1] "34 FALSE FALSE"
[1] "33 FALSE FALSE"
[1] "32 FALSE FALSE"
[1] "31 FALSE FALSE"
[1] "30 TRUE TRUE"
```

Here it was not 50/50, but at least you see that indeed it is a bit (deterministically) random.

So while this is definitely a usability issue (and ideally we would be able to solve that. I thought that maybe snapping would have helped, but that didn't seem to make a difference ..), I would for now just implement this as you were doing, and document this as a gotcha (also given that R `s2` is doing the same). And then we can later try to improve this.

Did you also test this example in BigQuery? (wondering if it can handle this case better ..)

For testing, we can use an example case where there is a shared boundary, but only fully shared edges (so there is never a vertex that falls somewhere in the middle of the edge of the other). For example:

```r
> parent_poly <- s2_make_polygon(c(-118, -118, -118, 34, 34, 34), c(60, 40, 23, 23, 40, 60))
> shared_bound_poly <- s2_make_polygon(c(-118, -118, 34, 34), c(40, 23, 23, 40))
> s2_contains(parent_poly, shared_bound_poly)
[1] TRUE
> s2_covers(parent_poly, shared_bound_poly)
[1] TRUE
```

JoelJaeschke:
Hey @jorisvandenbossche, thanks for your feedback.

> So while this is definitely a usability issue (and ideally we would be able to solve that. I thought that maybe snapping would have helped, but that didn't seem to make a difference ..), 

Behind the scenes, the `ApproxContains` function does use snapping with the given tolerance, if that is what you mean? Maybe, if this option is more exposed in higher-level APIs, that could be used? But in general, I agree that its probably easiest to just document this example and make the user aware.

> Did you also test this example in BigQuery? (wondering if it can handle this case better ..)

I did, BigQuery does in fact handle this case properly, although I only tested this with one polygon and ST_Covers. But happy to re-test some more cases.

> For testing, we can use an example case where there is a shared boundary, but only fully shared edges (so there is never a vertex that falls somewhere in the middle of the edge of the other).

Sounds good to me, I will push the changes

JoelJaeschke:
I implemented all the feedback from above and documented the behavior in the tests a bit. If there is anything you would like me to change, let me know.

After thinking about this more, there is probably one way how this could work properly. Given two `MutableS2ShapeIndex`  instances `a` and `b`, we would have to build the intersection of the two and then test the result for equality against `b`. If this is true, it should mean that `b` is fully covered by `a`.

I gave implementing this a go, but I struggle with how to make it work when just given two `MutableS2ShapeIndex` instances. Maybe someone who has more experience with s2geometry can take a stab at this.

tests/test_predicates.py
```diff
@@ -126,3 +127,90 @@ def test_disjoint():
...
+                # identical shared edges between polygons `a` and `b`, which does work as intended.
+                spherely.polygon(
+                    [(-118.0, 40.0), (-118.0, 23.0), (34.0, 23.0), (34.0, 40.0)]
+                ),
+                spherely.polygon(
+                    [(-117.0, 40.0), (-117.0, 35.0), (-115.0, 35.0), (-115.0, 40.0)]
+                ),
+            ]
+        ),
+        np.array([False, True, True, False, True, True, False, True, True]),
```
jorisvandenbossche, line 201: We don't really have extensive tests for `contains` with those corner case of on the boundary etc. Would it make sense to run this function for both, and so create a second numpy array with expected labels for `contains`? (that also makes it very explicit what the differences are, and whether we already implement the same nuance between contains vs covers or not)

tests/test_predicates.py
```diff
@@ -126,3 +127,90 @@ def test_disjoint():
...
+                spherely.linestring([(-120.0, 70.0), (-116.0, 37.0)]),
+                spherely.linestring([(-118.0, 41.0), (-118.0, 23.0)]),
+                spherely.linestring([(-117.0, 39.0), (-115.0, 37.0)]),
+                # Basic polygon test, crossing, shared boundary and interior
+                spherely.polygon(
+                    [(-120.0, 41.0), (-120.0, 35.0), (-115.0, 35.0), (-115.0, 41.0)]
+                ),
+                # TODO: This case is currently not fully correct. Supplying a polygon `a` and `b` and checking
+                # whether `a` covers `b`, where `a` and `b` share co-linear edges only works when the edges
+                # between them are identical. An example of this breaking would be a polygon `a` consisting of the edges AB, BC and CA and
```
jorisvandenbossche, line 188: Can you reduce the line length here for the comments? (that makes it a bit more readable)

jorisvandenbossche (commented):
Thanks a lot @JoelJaeschke! Added a few comments

src/predicates.cpp
```diff
@@ -117,4 +117,77 @@ void init_predicates(py::module& m) {
...
+
+              return s2geog::s2_intersects(a_index, b_index, closedOptions) &&
+                     !s2geog::s2_intersects(a_index, b_index, openOptions);
+          })),
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if A and B intersect, but their interiors do not intersect.
+        A and B have at least one point in common, where the common point
+        lies in at least one boundary.
```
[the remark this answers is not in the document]
JoelJaeschke: I think this is another one of these funny cases. Admittedly, I did not give much thought to this and the explanation is mostly copied from [ST_Touches](https://postgis.net/docs/ST_Touches.html) from PostGIS. I assume the reasoning in PostGIS to say 'in at least one boundary' is that points do not have a boundary, so for point/polygon or point/line tests, a point would have to lie on the boundary of the polygon/line, but would not be on its own boundary, if that makes sense? Happy to change the docstring though :D

tests/test_predicates.py
JoelJaeschke: Yeah, definitely makes sense to have this! I have refactored it slightly, although the new test for `contains` fails with what I believe are the correct outputs.

Currently, `contains` returns true for a point on the boundary of a polygon, however, it should not as there is no point that is common between the interior of the polygon and the point in question (https://postgis.net/docs/ST_Contains.html)

src/predicates.cpp
```diff
@@ -117,4 +117,77 @@ void init_predicates(py::module& m) {
...
+          py::vectorize(Predicate([](const s2geog::ShapeIndexGeography& a_index,
+                                     const s2geog::ShapeIndexGeography& b_index,
+                                     const S2BooleanOperation::Options& options) {
+              S2BooleanOperation::Options closedOptions = options;
+              closedOptions.set_polygon_model(S2BooleanOperation::PolygonModel::CLOSED);
+              closedOptions.set_polyline_model(S2BooleanOperation::PolylineModel::CLOSED);
+
+              S2BooleanOperation::Options openOptions = options;
+              openOptions.set_polygon_model(S2BooleanOperation::PolygonModel::OPEN);
+              openOptions.set_polyline_model(S2BooleanOperation::PolylineModel::OPEN);
```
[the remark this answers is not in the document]
JoelJaeschke: Agreed. However, especially in this case, do you think exposing `m_options` would be sufficient? I think we would actually need two `S2BooleanOperation::Options` objects with different 'settings' here to properly do the check.

Would it make sense to maybe move away from having the `Predicate` class and have something like one specialized struct per predicate? Considering also that there may be a few more edge cases and having a bit more flexibility in the implementation might make it easier to not affect other predicates?

tests/test_predicates.py
JoelJaeschke: Done

src/predicates.cpp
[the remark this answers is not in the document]
jorisvandenbossche: Ah yes, for this case having a separate class will work better (didn't think about the touches case doing two calls with different options). For the others it might work fine with the current Predicate structure.

But certainly feel free to add a separate class for TouchesPredicate and we can see from there

src/predicates.cpp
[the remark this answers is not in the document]
jorisvandenbossche: Ah, yes, for Point's that is true (and so the shapely version is actually not entirely accurate: https://shapely.readthedocs.io/en/2.0.2/reference/shapely.touches.html)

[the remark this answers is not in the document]
JoelJaeschke: Should we leave it as it is then?

src/predicates.cpp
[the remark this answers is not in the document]
JoelJaeschke: I have done so and exposed the options on the Predicate functor. The initialization should now happen only once.

JoelJaeschke:
Hey @jorisvandenbossche, is there anything missing that I should add and/or change here?

src/predicates.cpp
```diff
@@ -117,4 +148,64 @@ void init_predicates(py::module& m) {
...
     )pbdoc");
+
+    m.def("touches",
+          py::vectorize(TouchesPredicate()),
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if A and B intersect, but their interiors do not intersect.
+        A and B have at least one point in common, where the common point
+        lies in at least one boundary.
```
benbovy, line 159:
```suggestion
        Returns True if A and B intersect, but their interiors do not intersect.
        
        A and B must have at least one point in common, where the common point
        lies in at least one boundary.
```

src/predicates.cpp
```diff
@@ -117,4 +148,64 @@ void init_predicates(py::module& m) {
...
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if every point in A lies inside the interior or boundary of B.
+
+        Parameters
+        ----------
+        a, b : :py:class:`Geography` or array_like
+            Geography object(s)
+
```
benbovy, line 209:
```suggestion

        See Also
        --------
        covers
```

src/predicates.cpp
```diff
@@ -117,4 +148,64 @@ void init_predicates(py::module& m) {
...
+          py::arg("a"),
+          py::arg("b"),
+          R"pbdoc(
+        Returns True if every point in B lies inside the interior or boundary of A.
+
+        Parameters
+        ----------
+        a, b : :py:class:`Geography` or array_like
+            Geography object(s)
+
```
benbovy, line 189:
```suggestion

        Notes
        -----
        If A and B are both polygons and share co-linear edges,
        `covers` currently returns expected results only when those
        shared edges are identical.
 
```

Just to better notify users about this as it is not clear if / how easy it will be to work around it.

(an `Examples` section may better illustrate that behavior but it is something that we can address later more generally across the library)

benbovy (commented):
Thanks @JoelJaeschke for investigating this!

I just added some minor comments or suggestions regarding the docstrings. Overall this looks good to me!

jorisvandenbossche (approved):
Thanks a lot @JoelJaeschke!

jorisvandenbossche:
Given that this is something that ideally should be solved in the `s2geography` layer (since this also affects R), I opened an issue over there to describe this specific case -> https://github.com/paleolimbot/s2geography/issues/44 

So let's continue the discussion there.

PR #62, opened 2024-10-31
ENH: add boolean operations (overlays)

jorisvandenbossche:
WIP to add the overlay operations (union/intersection/(symmetric_)difference). 
They have a bunch of options that we should expose, and also need to add tests.


jorisvandenbossche:
Added some tests, so this is now the basics without any options (that can probably be added in follow-up PRs)

benbovy (approved):
Looks all good to me, thanks @jorisvandenbossche!

Issue #38, opened 2024-08-19
Add IO functions to read and write WKT and WKB


jorisvandenbossche:
We should have functions to create geographies from WKT and WKB, and also to convert to WKT and WKB.

(similarly to shapely's `from_wkt`/`from_wkb` and `to_wkt`/`to_wkb`)

At the moment `s2geography` does have a basic WKTWriter (which we already use in `__repr__`) and WKTReader, but proper (and easier) support probably depends on https://github.com/paleolimbot/s2geography/pull/16

benbovy:
Some naive questions as I'm not very familiar yet with (geo)arrow. 

IIUC https://github.com/paleolimbot/s2geography/pull/16 adds support for Arrow arrays so that we can easily reuse all the work done in geoarrow (and its implementation geoarrow-c) for parsing / encoding simple feature geometries.

- We want to support Numpy arrays for IO functions. So I guess we would need to internally convert Numpy <-> Arrow arrays to support that? How expensive would it be?
- Do we want direct support for Arrow arrays in Spherely IO functions? This could be useful I guess, even though other vectorized "universal" functions in Spherely would still support Numpy arrays only. Or maybe it would make sense to support Arrow arrays everywhere and internally convert to Numpy arrays when needed?
- Should we reuse geoarrow as well for coordinate operations (i.e., `get_coordinates`) to avoid reinventing the wheel, or would #37 be straightforward to extent to other geometries?

PR #68, opened 2024-11-10
ENH: add to_wkb / from_wkb

jorisvandenbossche:
Depends on https://github.com/paleolimbot/s2geography/pull/42

Closes https://github.com/benbovy/spherely/issues/38

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -46,7 +46,7 @@ Input/Output
 
    from_wkt
    to_wkt
-   fron_wkb
+   from_wkb
    to_wkb
    from_geoarrow
    to_geoarrow
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -46,6 +46,8 @@ Input/Output
 
    from_wkt
    to_wkt
+   fron_wkb
+   to_wkb
    from_geoarrow
    to_geoarrow
 
--- a/src/io.cpp
+++ b/src/io.cpp
@@ -54,6 +54,51 @@ private:
     std::shared_ptr<s2geog::WKTWriter> m_writer;
 };
 
+#if defined(S2GEOGRAPHY_VERSION_MAJOR) && \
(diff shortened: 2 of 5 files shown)


jorisvandenbossche:
The upstream change is merged in s2georaphy, so this should be ready now

docs/api.rst
```diff
@@ -46,6 +46,8 @@ Input/Output
 
    from_wkt
    to_wkt
+   fron_wkb
```
benbovy, line 49:
```suggestion
   from_wkb
```

src/io.cpp
```diff
@@ -54,6 +54,44 @@ class ToWKT {
...
+        }
+        m_reader = std::make_shared<s2geog::WKBReader>(options);
+    }
+
+    PyObjectGeography operator()(py::bytes a) const {
+        return make_py_geography(m_reader->ReadFeature(a));
+    }
+
+private:
+    std::shared_ptr<s2geog::WKBReader> m_reader;
```
benbovy, line 76: Use `unique_ptr` instead of `shared_ptr`? (Also in `ToWKB`) Unless I'm missing something?

benbovy (commented):
Thanks @jorisvandenbossche! Two minor comments below.

src/io.cpp
jorisvandenbossche: It might be that pybind11 requires this class to be copyable? (if you pass it as argument to vectorize) 
And that then gives compilation errors if using unique_ptr instead of shared_ptr

benbovy: > It might be that pybind11 requires this class to be copyable? (if you pass it as argument to vectorize)

Ah yes I see. Another option may be to add a `FromWKB` copy constructor (since this is what is passed to `py::vectorize`) although using a shared pointer is certainly fine.

jorisvandenbossche: I would personally leave it as is with a shared pointer then, to not make the small class more complicated (I don't think instantiation of the vectorized function is performance sensitive?)

PR #69, opened 2024-11-11
ENH: add projection support in geoarrow output

jorisvandenbossche:
Follow-up on https://github.com/benbovy/spherely/pull/53

### DIFF
--- a/src/geoarrow.cpp
+++ b/src/geoarrow.cpp
@@ -4,6 +4,7 @@
 #include "constants.hpp"
 #include "creation.hpp"
 #include "geography.hpp"
+#include "projections.hpp"
 #include "pybind11.hpp"
 
 namespace py = pybind11;
@@ -197,6 +198,7 @@ protected:
 
 ArrowArrayHolder to_geoarrow(py::array_t<PyObjectGeography> input,
                              py::object output_schema,
+                             Projection projection,
                              bool planar,
                              float tessellate_tolerance,
                              int precision) {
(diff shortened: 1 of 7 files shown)


jorisvandenbossche:
OK, rebased this on `main` after merging https://github.com/benbovy/spherely/pull/53, so this PR is now ready

src/projections.hpp
```diff
@@ -0,0 +1,38 @@
...
+    static Projection pseudo_mercator() {
+        return Projection(std::move(s2geog::pseudo_mercator()));
+    }
+    static Projection orthographic(double longitude, double latitude) {
+        return Projection(
+            std::move(s2geog::orthographic(S2LatLng::FromDegrees(latitude, longitude))));
+    }
+
+private:
+    std::shared_ptr<S2::Projection> m_s2_projection;
```
benbovy, line 35: Not sure why a shared pointer is needed (apart from that this is what is used in s2geography), but that's not a big deal anyway. Feel free to ignore this comment.

src/geoarrow.cpp
```diff
@@ -338,6 +340,10 @@ void init_geoarrow(py::module& m) {
             can be a ``pyarrow.DataType`` or ``pyarrow.Field``, and you can
             use the ``geoarrow.pyarrow`` package to construct such geoarrow
             extension types.
+        projection : spherely.Projection, default Projection.lnglat()
```
benbovy, line 343:
```suggestion
        projection : :py:class:`Projection`, default :py:meth:`Projection.lnglat`
```

Although maybe both are rendered as hyperlinks by sphinx? Either way it needs `Projection` be added to the API reference doc.

benbovy (commented):
Great! Just two small comments.

src/projections.hpp
jorisvandenbossche: We indeed need a shared pointer to pass to s2geography (and it is also what is returned by the s2geography constructors), so then I though that might be easier to just store the shared_ptr on the object, instead of unpacking it / packing it again in a shared_ptr ?

src/geoarrow.cpp
jorisvandenbossche: So I added it to the API docs to see if the linking would go automatically, which I suppose will be the case>? (at least for Geography it works) 

But our docs get build with the released s2geography, of course, so this is not yet included

benbovy: Yes let's figure out and address this later if needed when next s2geography version is released.

PR #71, opened 2024-11-13
Fix empty linestring creation

### LINKED ISSUE #67: Broken empty linestring repr with last s2geography dev version
Looks like the CI failing tests are caused by the changes in https://github.com/paleolimbot/s2geography/pull/37, which now raises an exception for empty S2Polyline. @jorisvandenbossche do you think we should fix that in s2geography or work around it in spherely to support empty linestrings?

benbovy:
Closes #67.

Instead of creating a PolylineGeography with an emtpy S2Polyline, create a PolylineGeography with no S2Polyline.


jorisvandenbossche (approved):
Thanks!

PR #77, opened 2024-11-18
ENH: Added length and perimeter implementation

JoelJaeschke:


src/accessors-geog.cpp
```diff
@@ -36,6 +36,14 @@ double area(PyObjectGeography a, double radius = EARTH_RADIUS_METERS) {
     return s2geog::s2_area(a.as_geog_ptr()->geog()) * radius * radius;
 }
 
+double length(PyObjectGeography a) {
+    return s2geog::s2_length(a.as_geog_ptr()->geog());
```
jorisvandenbossche, line 40: Also take into account the earth's radius like in `distance` above?

jorisvandenbossche (commented):
Thanks!

What's broken about the perimeter?

JoelJaeschke:
I am still getting used to thinking in spherical geometries. The test case I wrote for the perimeter function used a polygon that was rectangular in my head, so I expected its perimeter to be $4\pi / 180°$. However, not only this assumption was wrong (as angular distances between two points with the same latitude at higher latitudes are smaller than those near the equator), but I also did not take into account that all points at the poles are actually coincident, so my result was $3\pi/180°$ which I assumed to be some sort of bug.

Sorry for the confusion, everything was fine from the start!

jorisvandenbossche (approved):
Looks good!

Do you just want to add it to api.rst as well?

benbovy (commented):
Thanks @JoelJaeschke!

One question about the API: should be follow BigQuery's API or shapely's API? Or a mix of both? Some options:

1. Expose separate `length` and `perimeter` functions, with value zero returned for "incompatible" geography (like BigQuery, currently implemented in this PR)
2. Expose a unique `length` function (like Shapely)
3. Expose `length` and `perimeter` functions where `length` also returns the perimeter for an input Polygon.

benbovy:
> One question about the API: should we follow BigQuery's API or shapely's API? Or a mix of both?

Any thoughts on this @jorisvandenbossche?

jorisvandenbossche:
I think I like option 3: we provide a `length` that can handle both, and then still a `perimeter` for discoverability for people familiar with other systems like bigquery or postgis (and also `perimeter` is a more descriptive name in case of polygons).

(we should maybe consider adding a `perimeter()` in shapely as well)

I am also fine with leaving that for a follow-up, though, either way is fine.

JoelJaeschke:
I would be happy to change the PR accordingly if option 3 is the way forward!

benbovy (approved):
Let's choose the easy way and leave `length` and `perimeter` as it is in this PR for now.

Thanks @JoelJaeschke !

PR #79, opened 2024-11-24
ENH: add is_empty

jorisvandenbossche:
Another small function I could have used for plotting (filtering out the empties after calculating the intersection with the hemisphere)

I used the `s2geography` available function, but actually we should probably use the cached `is_empty()` method on our own Geography class? @benbovy


benbovy:
> but actually we should probably use the cached is_empty() method on our own Geography class?

Yes that might be a little more efficient but not a big difference I think. For consistency, we should probably use s2geography's function to set the cached value, actually.

benbovy:
Just added two commits that fix the conflicts with `main` and for consistency (API signatures and internal module organization).

I think it is good to go in! Thanks @jorisvandenbossche.

PR #81, opened 2024-12-02
Bump min s2geography version to 0.2.0

benbovy:
@jorisvandenbossche do you see any good reason to maintain compatibility with s2geography <0.2.0?

If not, I can remove the special cases (`#if` / `#endif` pre-processor blocks) for s2geography < 0.2.0 and I can move forward here with some refactoring of core functionality (e.g., clone s2geography objects) that would otherwise be cumbersome to make if we want to maintain compatibility with older s2geography versions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
     env:
       ABSL_VERSION: "20240722.0"
       S2GEOMETRY_VERSION: "0.11.1"
-      S2GEOGRAPHY_VERSION: "0.1.2"
+      S2GEOGRAPHY_VERSION: "0.2.0"
       CXX_STANDARD: 17
     strategy:
       fail-fast: false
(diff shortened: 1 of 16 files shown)


martinfleis:
Given spherely does not have any release compatible with 0.1.0 I don't see a reason to promise any backwards compatibility at this point.

benbovy:
Yes I agree. I also think it will be unlikely to have s2geography as a common dependency of multiple downstream packages in the same environment so maintaining backwards compatibility probably won't be very useful anyway except for development iterations, keeping CI running, etc. between the releases.

jorisvandenbossche:
Yep, fully agreed just bumping the minimum version, and cleaning up the code. 
Maybe you can just keep the machinery in cmakelists.txt to provide the version to be able to have those `#if` / `#endif` pre-processor blocks (we might use those in the near future again for temporary using things from s2geography 0.3dev until we do a new release)

jorisvandenbossche:
You can also remove some skips in the tests, eg 

https://github.com/benbovy/spherely/blob/d40a8d25fe6d18bd65d713cf9d5c658668c1a728/tests/test_io.py#L10-L13

benbovy:
This PR is ready for review. I'll implement new spherely features on top of s2geography 0.2.0 in follow-up PRs.

jorisvandenbossche (approved):

Issue #70, opened 2024-11-11
Exposing S2 / s2geography options (for boolean predicates / building operations)


jorisvandenbossche:
The spatial predicates already use those options, but at the moment we only set them under the hood and don't allow to override them by the user. In this case, `s2geography`'s functions directly allow you to specify the s2geometry [`S2BooleanOperation::Options`](https://github.com/google/s2geometry/blob/0fb1b8a1474f3137c8b44a7861e31f8efd1370e2/src/s2/s2boolean_operation.h#L281-L456)

Those `S2BooleanOperation` options allow you to set the snap function , the polygon/polyline model (open/closed), and a bunch of other things. 
And for example we implement `covers` by calling `contains` with a "closed" model.

The boolean (overlay, i.e. union/intersection/difference) operations I am adding in https://github.com/benbovy/spherely/pull/62 also have a somewhat similar (super)set of options. In this case, `s2geography` defines a [`GlobalOptions`](https://github.com/paleolimbot/s2geography/blob/21400cfc864203db7bb536f37bdbed3682e4140f/src/s2geography/build.h#L13-L35) options class (which includes the `S2BooleanOperation::Options`)  
Those options allow you, among other things, to specify whether the output should contain points/polylines/polygons (for example, allowing to limit the intersection of polygons to only return the polygonal intersections).

So for both the predicates and the overlays we need a way to allow the spherely user to customize those options.

---

In the R `s2` package, there is a `s2_options()` object that gathers all those options (https://r-spatial.github.io/s2/reference/s2_options.html) and that can be passed in those places (https://r-spatial.github.io/s2/reference/s2_contains.html, https://r-spatial.github.io/s2/reference/s2_boundary.html)

(the C++ side of this `s2_options()` object: https://github.com/r-spatial/s2/blob/main/src/s2-options.h)

jorisvandenbossche:
@paleolimbot based on your experience with R `s2` and `s2geography`, what is your current thought about the approach in the R package? 

Or is the `s2geography::GlobalOptions` already a kind of port of what you have in the R package to s2geography? 
It seems that effectively the R `GeographyOperationOptions` class is used (at the C++ level interacting with s2geometry) to get either the `s2geography::GlobalOptions` (in boolean operations) or the `S2BooleanOperation::Options` (in predicates)

paleolimbot:
The relevant R concept is `s2_options()`: https://r-spatial.github.io/s2/reference/s2_options.html

[...]

PR #82, opened 2024-12-05
ENH: add pickle support for Geography

### LINKED ISSUE #54: Support pickling Geography objects
At the moment we didn't implement this, so you will get `TypeError: cannot pickle 'spherely.Point' object`, but it would be useful to support it.

To provide a full exact roundtrip through pickling, we should maybe pickle the raw unit vector (x, y, z) of the `S2Point`, instead of going through latitude/longitude like other conversions (eg WKB), because there can always be some floating point difference in this conversion.

benbovy:
Closes #54.

A few comments:

- A Python tuple is used here to encode extra information that is not available or exposed from `s2geography::Geography` (i.e., the simple feature type and empty info). Alternatively, we could reuse s2geometry's Encoder / Decoder and pack everything into a single sequence of bytes, which is probably slightly more efficient, but I didn't want to mess with that and chose the easy way.

- Empty geography information is already handled internally by `s2geography::Geography::EncodeTagged` but it is not really available externally so encode it separately (that is a small amount of duplicated info, which is fine I guess).

- I thought it would be nice to also encode / decode the attached `geog_index()` if it exists, but s2geography decodes a `ShapeIndexGeography` as an `EncodedShapeIndexGeography` and I haven't figured out how we can reuse the latter (s2geography functions like predicates accept `ShapeIndexGeography`... I haven't seen yet how to convert between encoded shape index vs. shape index).

- For some reason direct equality check doesn't work with geography collections after pickle roundtrip (i.e., `roundtriped_collection == original_collection`) even though their WKT representation seem consistent. Not sure why.

### DIFF
--- a/tests/test_geography.py
+++ b/tests/test_geography.py
@@ -155,11 +155,11 @@ def test_equality() -> None:
 )
 def test_pickle_roundtrip(geog):
     pickled = pickle.dumps(geog)
-    roundtriped = pickle.loads(pickled)
+    roundtripped = pickle.loads(pickled)
 
-    assert spherely.get_type_id(roundtriped) == spherely.get_type_id(geog)
-    assert spherely.to_wkt(geog) == spherely.to_wkt(roundtriped)
+    assert spherely.get_type_id(roundtripped) == spherely.get_type_id(geog)
+    assert spherely.to_wkt(geog) == spherely.to_wkt(roundtripped)
 
     # FIXME: investigate why directy equality check doesn't work with collection
     if spherely.get_type_id(geog) != spherely.GeographyType.GEOMETRYCOLLECTION.value:
-        assert roundtriped == geog
+        assert roundtripped == geog
--- a/tests/test_geography.py
+++ b/tests/test_geography.py
@@ -1,3 +1,5 @@
+import pickle
+
 import pytest
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


tests/test_geography.py
```diff
@@ -123,3 +125,41 @@ def test_equality() -> None:
...
+        spherely.collection([spherely.point(40, 50)]),
+        # empty geography
+        spherely.point(),
+        spherely.linestring(),
+        spherely.polygon(),
+    ],
+)
+def test_pickle_roundtrip(geog):
+    pickled = pickle.dumps(geog)
+    roundtriped = pickle.loads(pickled)
```
jorisvandenbossche, line 158:
```suggestion
    roundtripped = pickle.loads(pickled)
```

jorisvandenbossche (approved):
Nice, looks good!

jorisvandenbossche:
> For some reason direct equality check doesn't work with geography collections after pickle roundtrip (i.e., `roundtriped_collection == original_collection`) even though their WKT representation seem consistent. Not sure why.

I have been thinking that we should add some kind of "low-level representation" of a Geography object as a kind of debug info (i.e. showing which underlying S2 objects are actually stored, what kind of loops, etc), to sometimes better understand how the object is actually represented (as the python repr using WKT is always a converted representation). Maybe something like that could to debug a case like this as well.

jorisvandenbossche:
Some possible lead, the `dimensions` and `nshape` attributes are different:

```python
In [17]: collection = spherely.collection([spherely.point(40, 50)])

In [18]: roundtripped = pickle.loads(pickle.dumps(collection))

In [19]: collection.nshape
Out[19]: 1

In [21]: roundtripped.nshape
Out[21]: 0

In [22]: collection.dimensions
Out[22]: 0

In [23]: roundtripped.dimensions
Out[23]: -1
```

benbovy:
Good catch! Yes this looks like a bug in s2geography, where `GeographyCollection::Decode` does not set the `num_shapes_` and `total_shapes_` internal members.

I'll open an new issue on the s2geography repo and push a temporary fix here.

benbovy:
Added a quick and dirty fix for the collection issue (clone the instance during decode so the returned object is created via the usual `s2geography::GeographyCollection` constructor that correctly updates internal members).

>  thought it would be nice to also encode / decode the attached geog_index() if it exists

This is nice to have but not really important at this point.

Ready to merge I think.

benbovy:
(note: it will probably be useful to later expose some `s2geography::EncodeOptions` in spherely as global options - or under context management, like suggested in #70)

benbovy:
> I have been thinking that we should add some kind of "low-level representation" of a Geography object as a kind of debug info (i.e. showing which underlying S2 objects are actually stored, what kind of loops, etc)

Yes that would be nice. Additionally, it would be nice to have some testing helpers like `spherely.testing.assert_geography_equals(a, b)` or use the `pytest_assertrepr_compare` hook to output more introspection results when the check fails.

jorisvandenbossche (approved):

PR #84, opened 2024-12-06
Misc. typing fixes

benbovy:
Many typing issues were not caught by mypy (untyped unit tests).

### DIFF
--- a/src/spherely.pyi
+++ b/src/spherely.pyi
@@ -299,7 +299,9 @@ class ArrowArrayExportable(Protocol):
     ) -> tuple[object, object]: ...
 
 class ArrowArrayHolder:
-    def __arrow_c_array__(self) -> tuple[object, object]: ...
+    def __arrow_c_array__(
+        self, requested_schema: object | None = None
+    ) -> tuple[object, object]: ...
 
 def to_geoarrow(
     input: Geography | T_NDArray_Geography,
--- a/src/spherely.pyi
+++ b/src/spherely.pyi
@@ -7,7 +7,6 @@ from typing import (
     Literal,
     Protocol,
     Sequence,
-    Tuple,
     TypeVar,
     overload,
 )
@@ -66,7 +65,7 @@ class Projection:
     @staticmethod
     def lnglat() -> Projection: ...
     @staticmethod
-    def speudo_mercator() -> Projection: ...
+    def pseudo_mercator() -> Projection: ...
     @staticmethod
     def orthographic(longitude: float, latitude: float) -> Projection: ...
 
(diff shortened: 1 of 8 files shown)


src/spherely.pyi
```diff
@@ -234,42 +251,61 @@ boundary: _VFunc_Nin1_Nout1[Literal["boundary"], Geography, Geography]
...
     def __arrow_c_schema__(self) -> object: ...
 
 class ArrowArrayExportable(Protocol):
     def __arrow_c_array__(
         self, requested_schema: object | None = None
-    ) -> Tuple[object, object]: ...
+    ) -> tuple[object, object]: ...
+
+class ArrowArrayHolder:
+    def __arrow_c_array__(self) -> tuple[object, object]: ...
```
jorisvandenbossche, line 302: This looks like a bug then in the implementation. It should match the `ArrowArrayExportable` above

jorisvandenbossche: OK, looking at the code, I wrote it to accept `*args, **kwargs` (I remember now that I struggled with making this more explicit keyword names),

So at least we should add that here to the signature as well. But ideally it is just the same as `ArrowArrayExportable` above (because that is the signature it should be in practice and that is useful for users to see)

benbovy: Ah yes I omitted `requested_schema` since the implementation returns an error when trying to pass anything other than the default value, but I'll update the signature since it still accepts the default value.

src/spherely.pyi
```diff
@@ -234,42 +251,63 @@ boundary: _VFunc_Nin1_Nout1[Literal["boundary"], Geography, Geography]
...
 class ArrowSchemaExportable(Protocol):
     def __arrow_c_schema__(self) -> object: ...
 
 class ArrowArrayExportable(Protocol):
     def __arrow_c_array__(
         self, requested_schema: object | None = None
-    ) -> Tuple[object, object]: ...
+    ) -> tuple[object, object]: ...
+
+class ArrowArrayHolder:
```
jorisvandenbossche, line 301: Given this is now the same as above, does it work in a pyi file to simply do:

```suggestion
class ArrowArrayHolder(ArrowArrayExportable): ...
```

?

benbovy:
I have no idea why CI passed in [20a5836](https://github.com/benbovy/spherely/pull/84/commits/20a5836050f0d1b917db71b141ba9d5dcc9f90ed) and not in [3ad4104](https://github.com/benbovy/spherely/pull/84/commits/3ad41041eacb16a673f53954c5883c6247e175c6) (both using mypy version <1). Anyway, using mypy 1.x seems to work in the last commit (I've run the workflow multiple times to make sure it is not flaky).

jorisvandenbossche (approved):

PR #87, opened 2024-12-10
Fix and rename: get_dimensions()

### LINKED ISSUE #57: Dimension property: returned value for geography collections
In the case where a collection contains features of different dimensions, the default value returned via s2geography (-1) is not consistent with the value returned by shapely (max dimension value found).

Which one do we want?

_Originally posted by @jorisvandenbossche in https://github.com/benbovy/spherely/pull/51#discussion_r1806473111_

benbovy:
Closes #57.

- Return values that are consistent with shapely for collections.
- Rename `get_dimensions` to `get_dimension`
- Remove properties `Geography.dimensions` and `Geography.nshape`.


benbovy:
Note: the `Geography.dimensions` property still returns the results from the underlying `s2geography::Geography::dimension()` so it is inconsistent with `spherely.get_dimensions()`.

I wonder if we shouldn't remove the Geography `dimensions` and `nshape` altogether...

benbovy:
Sorry it wasn't clear, I meant remove those properties from the Python public API.

(And maybe expose another function to besides `get_dimensions()` to expose both behaviors?)

jorisvandenbossche:
I think it would indeed be good to remove that attribute for now, because it probably will only be confusing (and we can always add it later again if we want to do so)

FWIW, I think the name of this function should actually be `get_dimension`, i.e. without the last `s` (not plural). It's how this is in shapely (and so there is an argument for consistency with shapely), but it seems a misnomer there, because all other similar functions use a singular term (`get_type_id`, `get_coordinate_dimension`, etc)

benbovy:
> I think the name of this function should actually be get_dimension, i.e. without the last s (not plural).

Agreed. non-plural form makes more sense and is used in R's s2 and bigquery.

Maybe we could have a temporary alias here for compatibility with Shapely? And at the same time try to suggest changing the name of the function in shapely (with some deprecation cycle)?

jorisvandenbossche (approved):

PR #89, opened 2024-12-10
DOC: API reference tweaks & clean-up

benbovy:
Mostly cosmetic changes to make the API reference documentation nicer and more consistent across the library.

More details in each commit message.

Also includes changes suggested in #87.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,6 +20,8 @@ intersphinx_mapping = {
     "shapely": ("https://shapely.readthedocs.io/en/latest/", None),
 }
 
+# TODO: seem to have no effect
+# (see https://github.com/sphinx-doc/sphinx/issues/10235)
 autodoc_typehints = "none"
 
 napoleon_google_docstring = False
@@ -28,8 +30,17 @@ napoleon_use_param = False
 napoleon_use_rtype = False
 napoleon_preprocess_types = True
 napoleon_type_aliases = {
+    # general terms
+    "sequence": ":term:`sequence`",
+    "iterable": ":term:`iterable`",
+    # numpy terms
     "array_like": ":term:`array_like`",
     "array-like": ":term:`array-like <array_like>`",
+    # objects without namespace: spherely
+    "EARTH_RADIUS_METERS": "spherely.EARTH_RADIUS_METERS",
+    # objects without namespace: numpy
+    "ndarray": "~numpy.ndarray",
+    "array": ":term:`array`",
 }
 
 # source_suffix = ['.rst', '.md']
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,8 +61,6 @@ html_theme_options = dict(
     use_repository_button=True,
     use_issues_button=True,
     home_page_in_toc=False,
-    extra_navbar="",
-    navbar_footer_text="",
 )
 
 html_static_path = ["_static"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,10 +21,6 @@ intersphinx_mapping = {
     "pyarrow": ("https://arrow.apache.org/docs/", None),
 }
 
-# TODO: seem to have no effect
-# (see https://github.com/sphinx-doc/sphinx/issues/10235)
-autodoc_typehints = "none"
-
 napoleon_google_docstring = False
 napoleon_numpy_docstring = True
 napoleon_use_param = False
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,6 +18,7 @@ intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "numpy": ("https://numpy.org/doc/stable", None),
     "shapely": ("https://shapely.readthedocs.io/en/latest/", None),
+    "pyarrow": ("https://arrow.apache.org/docs/", None),
 }
 
 # TODO: seem to have no effect
--- a/src/accessors-geog.cpp
+++ b/src/accessors-geog.cpp
@@ -63,12 +63,16 @@ double perimeter(PyObjectGeography a, double radius = numeric_constants::EARTH_R
 }
 
 void init_accessors(py::module& m) {
+    py::options options;
+    options.disable_function_signatures();
+
     m.attr("EARTH_RADIUS_METERS") = py::float_(numeric_constants::EARTH_RADIUS_METERS);
 
     m.def("centroid",
           py::vectorize(&centroid),
           py::arg("a"),
-          R"pbdoc(
+          R"pbdoc(centroid(a)
+
         Computes the centroid of each geography.
 
         Parameters
@@ -76,12 +80,18 @@ void init_accessors(py::module& m) {
         a : :py:class:`Geography` or array_like
             Geography object
 
+        Returns
+        -------
+        Geography or array
+            A single or an array of POINT geography object(s).
+
     )pbdoc");
 
     m.def("boundary",
           py::vectorize(&boundary),
           py::arg("a"),
-          R"pbdoc(
+          R"pbdoc(boundary(a)
+
         Computes the boundary of each geography.
 
         Parameters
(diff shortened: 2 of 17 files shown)


docs/conf.py
```diff
@@ -18,8 +18,11 @@
     "python": ("https://docs.python.org/3", None),
     "numpy": ("https://numpy.org/doc/stable", None),
     "shapely": ("https://shapely.readthedocs.io/en/latest/", None),
+    "pyarrow": ("https://arrow.apache.org/docs/", None),
 }
 
+# TODO: seem to have no effect
+# (see https://github.com/sphinx-doc/sphinx/issues/10235)
```
jorisvandenbossche, line 25: This comment is now no longer up to date? 
(the commit message in https://github.com/benbovy/spherely/pull/89/commits/c59675e82d612d9bcadd677540dd260a4c7ad9ee seemed to indicate that you changed to manual signatures so `autodoc_typehints = "none"` would actually work?)

src/geoarrow.cpp
```diff
@@ -255,20 +255,25 @@ ArrowArrayHolder to_geoarrow(py::array_t<PyObjectGeography> input,
...
+        &from_geoarrow,
+        py::arg("geographies"),
+        py::pos_only(),
+        py::kw_only(),
+        py::arg("oriented") = false,
+        py::arg("planar") = false,
+        py::arg("tessellate_tolerance") = 100.0,
+        py::arg("projection") = Projection::lnglat(),
+        py::arg("geometry_encoding") = py::none(),
+        R"pbdoc(from_geoarrow(geographies, /, *, oriented=False, planar=False, tessellate_tolerance=100.0, projection=Projection.lnglat(), geometry_encoding=None)
```
jorisvandenbossche, line 275:
```suggestion
        R"pbdoc(from_geoarrow(geographies, /, *, oriented=False, planar=False, tessellate_tolerance=100.0, projection=spherely.Projection.lnglat(), geometry_encoding=None)
```
?

src/geoarrow.cpp
```diff
@@ -315,28 +320,36 @@ void init_geoarrow(py::module& m) {
...
+        &to_geoarrow,
+        py::arg("geographies"),
+        py::pos_only(),
+        py::kw_only(),
+        py::arg("output_schema") = py::none(),
+        py::arg("projection") = Projection::lnglat(),
+        py::arg("planar") = false,
+        py::arg("tessellate_tolerance") = 100.0,
+        py::arg("precision") = 6,
+        R"pbdoc(to_geoarrow(geographies, /, *, output_schema=None, projection=Projection.lnglat(), planar=False, tessellate_tolerance=100.0, precision=6)
```
jorisvandenbossche, line 342:
```suggestion
        R"pbdoc(to_geoarrow(geographies, /, *, output_schema=None, projection=spherely.Projection.lnglat(), planar=False, tessellate_tolerance=100.0, precision=6)
```

src/geography.cpp
```diff
@@ -318,36 +345,92 @@ void init_geography(py::module &m) {
...
           py::vectorize(&get_type_id),
           py::arg("geography"),
-          R"pbdoc(
-        Returns the type ID of a geography.
+          R"pbdoc(get_type_id(geography)
 
-        - None (missing) is -1
-        - POINT is 0
-        - LINESTRING is 1
+        Returns the type ID of a geography.
```
jorisvandenbossche, line 358: I would personally keep this list here (but then expanded to be correct). Of course that gives two places we have to keep this up-to-date (although hopefully we won't change those values often after 0.1.0 ..). 
When using this function, it's nice to directly see what the return values mean without having to first check another docstring (when using the online documentation it is of course only a single click, but when in an interactive python session it is a bit more convoluted)

Or at least I would also mention ```:py:class:`GeographyType` ``` here already and not only in the returns section?

src/accessors-geog.cpp
```diff
@@ -63,70 +47,66 @@ double perimeter(PyObjectGeography a, double radius = numeric_constants::EARTH_R
 }
 
 void init_accessors(py::module& m) {
+    py::options options;
+    options.disable_function_signatures();
```
jorisvandenbossche, line 51: It is not possible to set this globally for all modules that will be added?

src/geography.cpp
```diff
@@ -410,7 +496,12 @@ void init_geography(py::module &m) {
         Parameters
         ----------
         geography : :py:class:`Geography` or array_like
-            Geography object(s)
+            Geography object(s).
+
+        Returns
+        -------
+        prepared : Geography or array
+            The same input Geography object(s) with an attached index.
```
jorisvandenbossche, line 504: When adding this, it might be worth explicitly mentioning in the docstring higher up that the function also updates the geographies in place, and not only the ones that are returned

src/predicates.cpp
```diff
@@ -65,27 +65,37 @@ class TouchesPredicate {
...
         Parameters
         ----------
         a, b : :py:class:`Geography` or array_like
-            Geography object(s)
+            Geography object(s).
+
+        Returns
+        -------
+        bool or array
+            Results of the ``intersects`` test.
```
jorisvandenbossche, line 89: Sidenote, and this is totally fine since you already added it! But personally I don't find it needed to strictly always have an _explanation_ of the result. It's good to know it returns a bool or boolean array, but that it is the result of `intersects` is IMO just duplicating what is already said above.

(thought the same about the `intersection` et al docstrings)

src/projections.cpp
```diff
@@ -7,12 +7,55 @@
...
 void init_projections(py::module& m) {
-    py::class_<Projection>(m, "Projection")
-        .def("lnglat", &Projection::lnglat)
-        .def("pseudo_mercator", &Projection::pseudo_mercator)
-        .def("orthographic", &Projection::orthographic);
+    py::class_<Projection> projection(m, "Projection", R"pbdoc(
+        Lightweight wrapper for selecting common reference systems used to
+        project Geography points or vertices.
+
+        Cannot be instanciated directly.
```
jorisvandenbossche, line 17:
```suggestion
        Cannot be instantiated directly.
```

src/geography.cpp
```diff
@@ -262,17 +283,23 @@ void init_geography(py::module &m) {
     // Geography types
 
     auto pygeography_types = py::enum_<GeographyType>(m, "GeographyType", py::arithmetic(), R"pbdoc(
-        The enumeration of Geography types
+        The enumeration of Geography types.
```
jorisvandenbossche, line 286: Not entirely sure why this happens, but for some reason this first line is "quoted" in the docstring, as if sphinx thinks it is indented (https://spherely--89.org.readthedocs.build/en/89/_api_generated/spherely.GeographyType.html)

src/accessors-geog.cpp
```diff
@@ -63,70 +47,66 @@ double perimeter(PyObjectGeography a, double radius = numeric_constants::EARTH_R
...
+
     m.attr("EARTH_RADIUS_METERS") = py::float_(numeric_constants::EARTH_RADIUS_METERS);
 
     m.def("centroid",
           py::vectorize(&centroid),
-          py::arg("a"),
-          R"pbdoc(
+          py::arg("geography"),
```
jorisvandenbossche, line 57: While we are at it (renaming those single args), should we make those first arguments positional-only? (so we avoid anyone ever doing `centroid(geography=...)` (renaming this keyword name is really only meant for being descriptive in the docs, I would say?)

jorisvandenbossche (commented):
Great work! 

I added some comments while going through the individual commits (thanks for the nice separate commits and messages, that certainly made it easier to review!), but you can see what you address here.

src/accessors-geog.cpp
benbovy: Yes probably better to set this once in `spherely.cpp` (the `options` lifetime would go through all `init_xxx` calls). I initially thought this would be set only for some functions or classes but I ended-up setting it for all functions.

docs/conf.py
benbovy: I think we can just remove `autodoc_typehints = "none"` now that we have switched to manual signatures globally.

The behavior is a bit confusing. IIUC when both `autodoc_docstring_signature=True` (default) *and* the 1st line of the docstrings looks like a signature, then Sphinx pops the 1st line and overrides the signature ignoring `autodoc_typehints` in the process (from https://github.com/sphinx-doc/sphinx/issues/10235#issuecomment-1059896282).

benbovy:
@jorisvandenbossche I addressed your review comments in the last two commits. Do you want to review it before merging this? Ortherwise this is ready I think.

jorisvandenbossche (approved):
Looks good, thanks for the updates!

PR #90, opened 2024-12-10
Align `GeographyType` values with `shapely.GeometryType`

### LINKED ISSUE #88: `spherely.GeographyType` vs. `shapely.GeometryType` id values
Because we don't have LINEARRING in spherely, the current id values differ between shapely geometries and spherely geographies starting from POLYGON. Should we offset the values in spherely so that we have consistent id values in both shapely and spherely?

benbovy:
Closes #88.

(This is on top of #89).


benbovy:
Any thoughts on this @jorisvandenbossche?

PR #95, opened 2025-02-04
Setup pixi

benbovy:
FYI @jorisvandenbossche.

There's no change in the existing functionality & configuration, I just added some extra configuration so it is possible to use [pixi](https://pixi.sh) for contributing to the development of Spherely. (I've recently switched to it for most of the projects I'm currently working on and I'm pretty happy with it!)

The only downside is that the dependencies are declared in more duplicated places, i.e., `pyproject.toml`, `ci/environment-xxx.yml`, `docs/environment.yml` files. Not a big deal I think, though.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,8 +30,79 @@ test = ["pytest>=6.0"]
 
 [tool.scikit-build]
 sdist.exclude = [".github"]
+build-dir = "build/{wheel_tag}"
 
 [tool.mypy]
 files = ["tests", "src/spherely.pyi"]
 show_error_codes = true
 warn_unused_ignores = true
+
+[tool.pixi.project]
+channels = ["conda-forge"]
+platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]
+
+[tool.pixi.environments]
+default = {features = [], solve-group = "default"}
+dev = {features = ["dev"], solve-group = "default"}
+test = {features = ["test", "dev"], solve-group = "default"}
+doc = {features = ["doc", "dev"], solve-group = "default"}
+lint = {features = ["lint", "dev"], solve-group = "default"}
+all = {features = ["lint", "doc", "test", "dev"], solve-group = "default"}
+
+[tool.pixi.dependencies]
+numpy = "*"
+
+[tool.pixi.feature.dev.dependencies]
+cxx-compiler = ">=1.9.0,<2"
+cmake = ">=3.31.5,<4"
+ninja = ">=1.12.1,<2"
+pybind11 = ">=2.13.6,<3"
+scikit-build-core = ">=0.10.7,<0.11"
+s2geometry = ">=0.11.1,<0.12"
+s2geography = ">=0.2.0,<0.3"
+
+[tool.pixi.feature.dev.pypi-dependencies]
+spherely = { path = ".", editable = true }
+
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -30,8 +30,79 @@ test = ["pytest>=6.0"]
...
+precommit-install = "pre-commit install"
+mypy = "python -m mypy"
+
+[tool.pixi.feature.doc.tasks]
+build-doc = "sphinx-build -M html docs docs/build"
+
+[tool.pixi.feature.dev.tasks.compile-commands]
+cmd = "cmake -GNinja -S. -Bbuild/compile-commands -DCMAKE_EXPORT_COMPILE_COMMANDS=ON && cp build/compile-commands/compile_commands.json ."
+inputs = ["CMakeLists.txt"]
+outputs = ["compile_commands.json"]
```
jorisvandenbossche, line 97: What is this one exactly doing?

pyproject.toml
```diff
@@ -30,8 +30,79 @@ test = ["pytest>=6.0"]
...
+[tool.pixi.feature.dev.tasks.configure]
+cmd = "cmake -GNinja -S. -Bbuild/python -DSKBUILD_PROJECT_NAME=spherely -DSKBUILD_PROJECT_VERSION=0.0.0"
+inputs = ["CMakeLists.txt"]
+outputs = ["build/python/CMakeFiles/"]
+
+[tool.pixi.feature.dev.tasks.compile]
+cmd = "cmake --build build/python --config Debug"
+depends-on = ["configure"]
+inputs = ["src/*"]
+outputs = ["build/python/spherely*"]
```
jorisvandenbossche, line 108: Are you using this essentially like a `python setup.py build_ext --inplace` when using setuptools, i.e. to just re-build the extension without going each time through a full `pip install ...` ?

jorisvandenbossche (commented):
Cool!

pyproject.toml
benbovy: The `compile_commands.json` file is commonly used by language servers like clangd and ccls to get project specific configuration (include paths, compiler flags...). A default search path for this file is the project's root directory.

All I need to do in my case (emacs + eglot + clangd) is run this task once (re-run it after adding new files) for getting linting, auto-completion, etc.

I think that vscode supports it too via the C++ extension, although by default the latter uses another way to get project configuration.

(Note: we could add a description to this pixi task)

pyproject.toml
benbovy: The idea here is compiling the Python module simply by calling cmake instead of via a Python build tool and the scikit-build-core backend, which is mostly for getting more readable compilation outputs (not hidden by the tool, no extra verbosity, colors, etc.) in case we need to inspect errors or warnings.

Pixi does not run or depend on this task for re-installing Spherely, I think it does that with `uv` since Spherely is declared as a local pypi-dependency. Spherely builds are configured with no isolation and in a defined build directory (so re-builds are faster).

jorisvandenbossche: Thanks for the explanation!

benbovy: That said, I had troubles with the spherely Python module not being automatically re-built / re-installed each time when doing `pixi install ...` or `pixi run tests` after changing the C++ code. Maybe I'm doing something wrong? https://github.com/prefix-dev/pixi/issues/1695 may be relevant too.

PR #98, opened 2025-02-04
Align GeographyType values with shapely.GeometryType

### LINKED ISSUE #88: `spherely.GeographyType` vs. `shapely.GeometryType` id values

benbovy:
Replaces #90 (automatically closed by deleting target branch != main).
Closes #88.

### DIFF
--- a/src/geography.cpp
+++ b/src/geography.cpp
@@ -285,15 +285,15 @@ void init_geography(py::module &m) {
     pygeography_types.value(
         "LINESTRING", GeographyType::LineString, "Single line geography type (1).");
     pygeography_types.value(
-        "POLYGON", GeographyType::Polygon, "Single polygon geography type (2).");
+        "POLYGON", GeographyType::Polygon, "Single polygon geography type (3).");
     pygeography_types.value(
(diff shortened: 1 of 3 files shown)


jorisvandenbossche (approved):
I don't have a strong opinion about whether it is better to align, or better to just have our own enum (or align with WKB values, which is 1-7). 
Whichever we choose, it should always be recommended to compare with the enum and not with a direct integer, and if you follow that then the actual number does not matter too much.

benbovy:
> Whichever we choose, it should always be recommended to compare with the enum and not with a direct integer, and if you follow that then the actual number does not matter too much.

I agree, however there's no easy way to support that here (https://github.com/benbovy/spherely/pull/51#discussion_r1797638051) so we have to stick with comparing values for now (hence this PR and #88).

That said, aligning with WKB is a good point (I didn't know about that) and probably better than aligning with shapely? I'll leave this PR and #88 open and move on with the 0.1.0 release.

Issue #52, opened 2024-10-08
More robust/flexible numpy dtype and ufuncs


benbovy:
Not sure that we need this now, but [here](https://github.com/EricCousineau-TRI/pybind11/commit/31fe60a00ca6cbdd83bc28044267022d8301d6a7#diff-341a738925956060860e5cde2f62466961ae52d2f0d8606f75ffc00932979186) is an implementation of Numpy user dtypes and ufuncs with pybind11.

jorisvandenbossche:
I also assume that it will require numpy>=2? 

It might certainly be nice to at some point use a custom user-defined numpy dtype, instead of the object dtype that can hold anything. 

But the current setup works quite OK I think (and you get a nice error message if some value in the object array is not a Geography), so I would also say to first focus on getting a minimal viable package with the current approach, and we can always improve this later.

benbovy:
> I also assume that it will require numpy>=2?

Not sure about that, the commit I linked above is from 6 years ago. I also don't really know whether there would be a more elegant, numpy 2 solution nowadays.

> It might certainly be nice to at some point use a custom user-defined numpy dtype, instead of the object dtype that can hold anything. But the current setup works quite OK I think

Yes definitely! Just wanted to link this here as a useful reference in case we might eventually need it. The ufunc feature is also relevant (perhaps more than user dtypes) to work around the limitations of `pybind11::vectorize`.

jorisvandenbossche:
> > I also assume that it will require numpy>=2?
> 
> Not sure about that, the commit I linked above is from 6 years ago. I also don't really know whether there would be a more elegant, numpy 2 solution nowadays.

Ah, I hadn't checked the date of the commit ;) 
Yes, nowadays numpy exposes a newer API to properly create custom dtypes (and expose those to the user in Python), so I assume the linked code might need to be updated for that.

I am also not entirely sure if the custom dtype already supports object dtype (I know when we explored an initial version of it two years ago for shapely, that wasn't yet the case)

PR #114, opened 2026-04-20
Add vectorized spherely.polygons() constructor

thodson-usgs:
## Summary

Adds `spherely.polygons(shells, oriented=False, validate=True)` — a numpy-native vectorized POLYGON constructor that accepts an `(N, K, 2)` array of shell coordinates and returns a `(N,)` object array of POLYGON Geographies.

This is the batch-construction counterpart to the existing `spherely.points(coords)` and is the lane for building many polygons of uniform shape (e.g. rectilinear grid cells) without per-polygon Python call overhead.

Motivated by workloads that construct large batches of polygons with uniform ring shape — e.g. spherical regridding over a lat/lon grid — where the Python-side call overhead dominates even though the geometry itself is trivial. Refs #52.

## What's in this PR

- `src/creation.cpp`
  - New `polygons(...)` function and `m.def("polygons", ...)` binding.
  - Extracted `make_s2loop_from_points(const std::vector<S2Point>&, check, oriented)` as the shared loop-construction core; the existing `make_s2loop` template now delegates to it. Callers drop the closed-ring duplicate vertex before calling (no silent mutation).
  - `make_s2polygon` gains an optional `check` parameter that gates the `S2Polygon::IsValid()` call. Default `true` preserves all existing behavior.
  - `polygons()` reuses a single `std::vector<S2Point>` scratch buffer across polygons so the per-ring vertex allocation is amortized.
- `src/spherely.pyi` — type stub.
- `tests/test_creation.py` — 6 tests covering shape/dtype, WKT equivalence with `create_polygon`, closed-ring auto-closing, `oriented=True`, shape errors, invalid-ring error path, and `validate=False` skipping validation.

## Limitations vs `create_polygon`

Called out explicitly in the docstring:

- **All shells share the same vertex count** (enforced by the numpy `(N, K, 2)` input shape).
- **No holes** — use `create_polygon` in a Python loop if you need them. Dropped per reviewer suggestion; the grid-cell use case this PR is built around doesn't need them, and the uniform-ndarray API is a bad fit for variable hole counts anyway.

## Performance

Measured on an M-series MacBook with rectilinear (K=4) quads, best of 5 runs:

|    n    | scalar loop | vec (validate=True) | vec (validate=False) |
|---------|-------------|---------------------|----------------------|
|  10,000 |    21.9 ms  |        17.1 ms      |         8.3 ms       |
| 100,000 |   229.2 ms  |       181.2 ms      |        92.5 ms       |
| 500,000 |  1202.7 ms  |       969.0 ms      |       496.9 ms       |

- **`validate=True`:** ~1.3× vs the scalar `[create_polygon(s) for s in shells]` loop.
- **`validate=False`:** ~2.5× vs scalar, ~2× vs `validate=True`. `S2Polygon::IsValid()` is roughly half the per-polygon cost; skipping it is safe when the input is valid by construction (grid cells, pre-validated geometries).

## Test plan

- [x] All existing tests pass (full `tests/test_creation.py` — 40/40).
- [x] New tests cover shape/dtype, WKT equivalence with `create_polygon` (folded into the basic test), orientation, closed rings, the invalid-ring error path, shape-error paths, and `validate=False`.
- [x] `black`, `clang-format`, pre-commit clean on changed files.
- [x] CI across Linux/Windows/macOS.

## Review feedback addressed (commit 587e494)

- [x] Renamed `check` → `validate` (`@jorisvandenbossche`).
- [x] Dropped the `holes` parameter and its implementation.
- [x] Docstring now spells out the uniform-vertex-count + no-holes limitations.
- [x] Merged the scalar-equivalence test into the basic test.

Perf numbers above are re-measured on the post-review commit; no change to the advertised speedups.

### DIFF
--- a/src/creation.cpp
+++ b/src/creation.cpp
@@ -65,22 +65,17 @@ std::vector<S2Point> make_s2points(const std::vector<V>& points) {
     return std::move(s2points);
 }
 
-// create a S2Loop from a pre-built vector of S2Point.
+// create a S2Loop from a pre-built vector of S2Point. Callers are responsible
+// for dropping any closed-ring duplicate vertex before calling.
 //
 // Normalization (to CCW order for identifying the loop interior) and validation
 // are both enabled by default.
 //
-// If the input loop is already closed, the duplicate last vertex is dropped.
-//
 // TODO: add option to skip normalization.
 //
-std::unique_ptr<S2Loop> make_s2loop_from_points(std::vector<S2Point> s2points,
+std::unique_ptr<S2Loop> make_s2loop_from_points(const std::vector<S2Point>& s2points,
                                                 bool check = true,
                                                 bool oriented = false) {
-    if (!s2points.empty() && s2points.front() == s2points.back()) {
-        s2points.pop_back();
-    }
-
     auto loop_ptr = std::make_unique<S2Loop>();
 
     loop_ptr->set_s2debug_override(S2Debug::DISABLE);
@@ -105,13 +100,18 @@ template <class V>
 std::unique_ptr<S2Loop> make_s2loop(const std::vector<V>& vertices,
                                     bool check = true,
                                     bool oriented = false) {
-    return make_s2loop_from_points(make_s2points(vertices), check, oriented);
+    auto s2points = make_s2points(vertices);
+    if (!s2points.empty() && s2points.front() == s2points.back()) {
+        s2points.pop_back();
+    }
+    return make_s2loop_from_points(s2points, check, oriented);
 }
 
 // create a S2Polygon.
 //
 std::unique_ptr<S2Polygon> make_s2polygon(std::vector<std::unique_ptr<S2Loop>> loops,
-                                          bool oriented = false) {
+                                          bool oriented = false,
+                                          bool check = true) {
     auto polygon_ptr = std::make_unique<S2Polygon>();
     polygon_ptr->set_s2debug_override(S2Debug::DISABLE);
 
@@ -122,7 +122,7 @@ std::unique_ptr<S2Polygon> make_s2polygon(std::vector<std::unique_ptr<S2Loop>> l
     }
 
     // Note: this also checks each loop of the polygon
-    if (!polygon_ptr->IsValid()) {
+    if (check && !polygon_ptr->IsValid()) {
         std::stringstream err;
         S2Error s2err;
         err << "polygon is not valid: ";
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
@thodson-usgs thanks a  lot for the PR!

This is certainly functionality we want (we might want to think about the exact API and naming, since we also have `create_polygon()`, although `polygons()` also make sense for consistency with shapely)

FWIW I think it would also be fine to leave out holes for now (if that is not something you need)

> **`check` default is `True`** so `polygons()` has the same safety guarantees as scalar `create_polygon`. `False` is explicit opt-in for valid-by-construction inputs. Happy to rename (`validate`? `trusted=False`?) if there's a preferred convention.

I think enabling validation by default is indeed the sensible default behaviour. For naming, I think I might like `validate=True/False` 
(I don't think we already expose this anywhere else, right? although it could probably also make sense for other creation methods)

> **`holes` representation** is a sequence of length N where each entry is either `None` or an `(H_i, K_h, 2)` array. This allows variable hole counts per polygon while keeping the common no-holes case cheap. Alternative API shapes welcome.

As mentioned above, I would also be fine with punting on holes for now. At least in shapely, I don't think that is used very often, but there we actually require holes to be an ndarray, so that makes that every polygon needs to have a uniform number of holes. The more flexible variant you added here is probably more useful, although I think the typical use case for `polygons` (like creating a grid) won't need it?

src/creation.cpp
```diff
@@ -513,4 +589,52 @@ void init_creation(py::module& m) {
...
+          py::arg("oriented") = false,
+          py::arg("check") = true,
+          R"pbdoc(polygons(shells, holes=None, oriented=False, check=True)
+
+        Create an array of polygons from a numpy array of ring coordinates.
+
+        This vectorized constructor is functionally equivalent to calling
+        :py:func:`create_polygon` for each shell but avoids the per-polygon
+        Python parsing overhead, making it much faster when building many
+        polygons of uniform shape (e.g. grid cells).
```
jorisvandenbossche, line 575: It probably makes sense here to mention to limitation compared to `create _polygon`, i.e. that this method is meant for creating size-uniform polygons (with a numpy ndarray input, all shells have the same number of vertices)

tests/test_creation.py
```diff
@@ -387,6 +388,144 @@ def test_multipolygon_invalid_geography() -> None:
...
+            [(0, 0), (2, 0), (2, 2), (0, 2)],
+            [(10, -5), (12, -5), (12, -3), (10, -3)],
+            [(-30, 40), (-28, 40), (-28, 42), (-30, 42)],
+        ],
+        dtype=np.float64,
+    )
+    vectorized = spherely.polygons(shells)
+    scalar = [spherely.create_polygon(ring.tolist()) for ring in shells]
+
+    assert [spherely.to_wkt(p) for p in vectorized] == [
```
jorisvandenbossche: You can include this verification with scalar result also in the test above (instead of putting that in a separate test)?

thodson-usgs:
Thanks for the feedback! I'm working with Claude to do some profiling on another project, and it honed in on a two issues in`spherely`.  I'll do my best to keep the changes tight and not bug you too much :)

tests/test_creation.py
thodson-usgs: will do

